PR #10, opened 2024-04-26
introduction session

sbfnk:
The idea for this is a self-coding session at the end of the first session ("R and statistical concepts used").

It includes a suggested teaching method for the rest of the course:
1. simulation (what process generates the kind of data I see)
2. initial estimation (using the simulation function to conduct inference)
3. full estimation (with stan)
4. open-source packages (what's out there to do this work)

step (2) might turn out not to be needed but is there to smooth the transition to stan.

Review on this PR please on (1) the general setup and (2) any errors/typos etc. More substantial changes to this session (which may well be needed/sensible) would be better in a separate PR, I think.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,6 +18,9 @@ URL: https://github.com/nfidd/nfidd
 BugReports: https://github.com/nfidd/nfiidd/issues
 Depends:
     R (>= 4.2.0)
+Suggests:
+    rstan,
+    fitdistrplus
 Encoding: UTF-8
 Language: en-GB
 LazyData: true
--- a/Rmd/introduction.Rmd
+++ b/Rmd/introduction.Rmd
@@ -1,0 +1,165 @@
+---
+title: "Introduction to the teaching method"
+---
+
+```{r seed, cache=FALSE, echo=FALSE}
+set.seed(1234)
+```
+
+# Objectives
+
+The aim of this session is to introduce both the concepts that will be used throughout the course as well as the teaching philosophy that it will follow.
+Throughout the course your learning will proceed in 4 steps:
+
+1. **simulation** to get an understanding of the process that we typically assume to generate observed infectious disease data
+2. **initial estimation** using standard R tools (especially the `optim()` function) as a way to link the simulation process to observations
+3. **full estimation** using the statistical programming language [stan](https://mc-stan.org/) as a way to obtain a range of flexible estimates with uncertainty 
+4. **open-source tools** that implement the methods encountered and allow you to avoid having to write your own models.
+
+We will illustrate this approach using a simple non-epidemiological example before moving on to infectious diseases in the next session.
+We will investigate a scenario where we're observing coin flips but we don't know if the coin is biased towards one side or the other.
+
+# Simulating coin flips
+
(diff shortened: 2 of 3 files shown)


DESCRIPTION
```diff
@@ -18,6 +18,9 @@ URL: https://github.com/nfidd/nfidd
 BugReports: https://github.com/nfidd/nfiidd/issues
 Depends:
     R (>= 4.2.0)
+Suggests:
+    rstan,
```
seabbs, line 22: Do we want to do this to ourselves and our course participants - i.e maybe we bite the small compilation bullet and use `cmdstanr`? The advantage of this potentially is that if they can't install `cmdstan` as a dependency they can still interact with all the material as installing deps won't fail for them?

Rmd/introduction.Rmd
```diff
@@ -0,0 +1,165 @@
...
+Let us assume we are completely ignorant, i.e. $p$ is equally likely to be any value between 0 and 1.
+In that case, we can use a _uniform prior_ bounded by 0 and 1, or in mathematical notation
+
+$$
+p \sim \mathcal{U}(0, 1)
+$$
+
+This is just another way of saying that any value of $p$ between 0 and 1 has the same _prior probability density_.
+
+When doing statistical inference we're often better off using distributions that don't have sharp edges like the uniform distribution.
```
seabbs, line 101:
```suggestion
When doing statistical inference we're often better off using distributions that don't have sharp edges like the uniform distribution and that better reflect our understanding of the world.
```

Rmd/introduction.Rmd
```diff
@@ -0,0 +1,165 @@
...
+In this model, we have defined three sections:
+- `data` represents all the things that aren't estimated, i.e. the things before the bar (|) in the likelihood function
+- `parameters` are the things that are estimated
+- `model` defines the prior and likelihood
+
+In order to conduct inference, we use the `sampling()` function in the `rstan` package
+
+```{r}
+library("rstan")
+model <- stan_model(model_code = coin_model) ## this may take a little while
```
seabbs, line 137: set cores?

Rmd/introduction.Rmd
```diff
@@ -0,0 +1,165 @@
...
+```{r}
+stan_result
+```
+
+**Take 15 minutes*** to experiment with the code above: how does the error of the estimate change with the number of trials, or the given `k_heads`.
+Does the result match what you would expect?
+You could change the prior distribution for the bias to a uniform distribution by changing the `beta` distribution above to a `uniform(lower_bound, upper_bound)` (replacing `lower_bound` and `upper_bound` with the appropriate parameters). 
+Does this change your results?
+
+# Open-source tools
```
seabbs, line 152: Do we want to give some tools for interacting with stan models? i.e `tidybayes` and `ggdist`?

seabbs (approved):
This is really nice and reads well. I think we can discuss a few details and decide to move them to issues. 

I generally like the framework and it seems like a good one for the rest of the course.

DESCRIPTION
sbfnk: Ah yes that makes sense if we don't ship precompiled models (which we probably don't want).

Rmd/introduction.Rmd
sbfnk: I was thinking of this section as "open-source tools that achieve what we're trying to do here" - but yes, we'll probably want to introduce these at some point somewhere.

PR #14, opened 2024-04-30
Further modularise learning objectives

seabbs:
Thinking about the learning objectives as a lesson plan I have further modularised them and made a distinct sections for biases in delay distributions, ensemble modelling, and delay estimation.

The biases in delay distribution section I think particularly makes sense as then it is clear how to pull things from our two recent papers on this.

### DIFF
--- a/learning_objectives.md
+++ b/learning_objectives.md
@@ -12,17 +12,14 @@
 - understanding of the ubiquity of delays in epidemiological data
 - understanding of how delays affect population-level epidemiological data via discrete convolutions
 - ability to apply convolutions of discrete probability distributions to epidemiological data in R
+
+## Biases in delay distributions
+
 - understanding of how censoring affects the estimation and interpretation of epidemiological delay distributions
 - ability to estimate parameters of probability distributions from observed delays, taking into account censoring, using R
-
-## Right truncation and nowcasting
-
 - understanding of right truncation in epidemiolgical data
 - ability to estimate parameters of probability distributions from observed delays, taking into account truncation, in R
-- understanding of nowcasting as a particular right truncation problem
-- ability to perform a simple nowcast in R
-- awareness of the breadth of methods to perform nowcasting
-
+  
 ## R estimation and the renewal equation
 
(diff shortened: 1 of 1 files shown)


learning_objectives.md
```diff
@@ -32,13 +29,23 @@
...
+- R estimation as a nowcasting problem
+
 ## Forecasting
 
 - understanding of forecasting as an epidemiological problem, and its relationship with nowcasting and R estimation
 - understanding of the difference between forecasts, projections and scenarios
 - familiarity with common forecasting models and their properties, and applicability in epidemiology
 - ability to use a common forecasting model on an epidemiological time series in R
-- ability to use a mechanistic model for forecasting an epidemiological time series in R
+- ability to use a semi-mechanistic model for forecasting an epidemiological time series in R
```
seabbs, line 45: I changed this as I assumed we would actually extend the renewal model to be used for forecasting.

sbfnk: it might be nice to look at the whole breadth, both mechanistic + semi-mechanistic, if we have the time. Where mechanistic ones generally don't have a cone of uncertainty so could discuss impact of strong model prior etc.

sbfnk (approved):
Thanks, makes sense.

sbfnk:
@seabbs what was the rationale for moving nowcasting to after R estimation? To me conceptually it flows more naturally to have nowcasting first as particular case of data right truncation and not requiring introduction of latent processes.

seabbs:
The thought was that for most nowcasting you need a generative model for what you would eventually observe and following on from @adrian-lison nice paper that should really be one that reflects your understanding of the generative process. 

If we do nowcasting after renewal we can teach that as we can combine the previous two bits (delays and renewal) into the joint model of them both (even if only in a restricted case like EpiNow2)

If we do it before then we either have to be very confusing or not try and teach above point.

seabbs:
(and as a more abstract point I think it's useful to be clear that everything ideally requires a latent process - it's just that sometimes we can simplify it away with assumptions)

sbfnk:
Ah yes that makes sense.

PR #15, opened 2024-04-30
split learning objectives into sessions

sbfnk:
~This is already superseded by #14 but were my initial thoughts for splitting it into 5 session of 3 hours. To be closed and replaced by a PR that introduces a schdule.~ superseded below


sbfnk:
Now replaced by a tentative (tight!) schedule. We could and perhaps should still consider cutting some material.

sbfnk:
Will need to update to https://github.com/nfidd/nfidd/pull/14#issuecomment-2092784489 - will do following review (which I suspect may end up in restructuring anyway).

schedule.md
```diff
@@ -0,0 +1,89 @@
+# Session plan
+
+## Session 1: R and statistical concepts used
+
+Some of this could be pre-course reading material, see e.g. https://sbfnk.github.io/mfiidd/slides/rintro.pdf
+
+Monday 9:30-11:00
+
+- Introduction to R concepts used in the course (15 mins)
```
seabbs, line 9: Can we really get away without a brief intro to stan itself?

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- Introduction to statistical concepts used in the course and what they look like in stan (30 mins)
+- Practice session: probability distribution and estimates (30 mins)
```
seabbs, line 11: I think we could kill this and just have it be part of session 2?

seabbs: if we did that it could give us back 30 minutes to have a short lecture on censoring and another on truncation in the following sessions.

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- Wrapup (15 mins)
+
+## Session 2: Delay distributions
```
seabbs, line 14: Personally, I would favour splitting this as proposed in the learning objects (i.e delays very simply and then biases (i.e censoring and truncation).

seabbs: That would also help as it would make it clear they are general biases one can see and so maybe help people in other areas of their modelling lives

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- Practice session: nowcasting in R (45 mins)
+- Wrapup (15 mins)
+
+## Session 4: R estimation and the renewal equation
+
+Tuesday 09:00-10:30
+
+- Introduction to the time-varying reproduction number (20 mins)
+- Practice session: using the renewal equation to estimate R (45 mins)
+- Practice session: combining nowcasting and R estimation (25 mins)
```
seabbs, line 41:
```suggestion
- Practice session: Combining R estimation and delay distributions
```

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Session 5: Forecasting concepts
+
+Tuesday 11:00-12:30
+
+- Wrap-up of previous session (15 mins)
+- Introduction to forecasting as an epidemiological problem, and its relationship with nowcasting and R estimation (10 mins)
+- Introduction to forecast evaluation (20 mins)
+- Practice session: evaluate the nowcasts from session 3b (45 mins)
```
seabbs, line 50: before we talk about eval can we give people 15 minutes to play with the model we will eval to make some forecasts (but not use these in the eval session?)

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- Wrapup (15 mins)
+
+## Session 2: Delay distributions
+
+Monday 11:00-12:30
+
+- Introduction to epidemiological delays and how to represent them with probability distributions (15 mins)
+- Practice session: simulate and estimate epidemiological delays (30 mins)
+- Practice session: introducing daily censoring (30 mins)
+- Wrapup (15 mins)
```
seabbs, line 21: All the early sessions havee wrap up sections and none of the later ones do. I think its a nice thing to have but maybe all should have them and we should aim for no more than 10 minutes?

schedule.md
```diff
@@ -0,0 +1,89 @@
+# Session plan
+
+## Session 1: R and statistical concepts used
+
+Some of this could be pre-course reading material, see e.g. https://sbfnk.github.io/mfiidd/slides/rintro.pdf
```
seabbs, line 5: I can make some stan precourse material but I think we will still want a 15 minute review lecture of what it is.

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- Wrap-up of previous session (15 mins)
+- Introduction to forecasting as an epidemiological problem, and its relationship with nowcasting and R estimation (10 mins)
+- Introduction to forecast evaluation (20 mins)
+- Practice session: evaluate the nowcasts from session 3b (45 mins)
+
+## Session 6: Forecasting models
+
+Tuesday 14:00-15:30
+
+Split group in 3 and get students to work together on using one of
```
seabbs, line 56: I see why you want to do this (as it leads to diverse models for ensembling) but I'm not sure it won't be too much vs just building on the model from the rest of the course.

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+Tuesday 11:00-12:30
+
+- Wrap-up of previous session (15 mins)
+- Introduction to forecasting as an epidemiological problem, and its relationship with nowcasting and R estimation (10 mins)
+- Introduction to forecast evaluation (20 mins)
+- Practice session: evaluate the nowcasts from session 3b (45 mins)
+
+## Session 6: Forecasting models
+
+Tuesday 14:00-15:30
```
seabbs, line 54: Are these days correct?

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+- a time-series model
+- a semi-mechanistic model
+- a mechanistic model
+to produce nowcasts/forecasts (90 mins)
+
+## Session 7: Evaluating forecasts
+
+Wednesday 16:00-17:30
+
+- Practice session: evaluate the forecasts from session 6 (30 mins)
```
seabbs, line 66: What happens if they don't get anything working? How do we combine everyone's forecasts for them to evaluate? I think I would rather we make them some forecasts that are related to the ones they make but not depend on those. Could then stretch to have them compare our forecasts with their own?

schedule.md
```diff
@@ -0,0 +1,89 @@
...
+Wednesday 09:00-10:30
+
+- Practice session: starting from a linelist, produce:
+  - nowcasts
+  - R estimation
+  - forecasts
+  using the tools created in this course (60 mins)
+- Discussion (30 mins)
+
+## Session 9: Wrap-up and overview of available tools
```
seabbs, line 82: I think we should drop this and give more space to other sessions. We can have post course reading for other tools etc.

seabbs: and or make it an extension session for the very good. I think most people will not be able to do session 8 in an hour.

seabbs:
> (tight!) schedule.

Currently feels a bit like Hermoine's class schedule from book 3 (due to the time travel)

schedule.md
sbfnk: No, good point.

schedule.md
sbfnk: Agreed

schedule.md
sbfnk: > Personally, I would favour splitting this as proposed in the learning objects (i.e delays very simply and then biases (i.e censoring and truncation).

Definitely, that's the idea - just not being clear in the text perhaps.

schedule.md
sbfnk: I went back and forth on this. Is it worth first introducing what we're aiming for in forecast or first play with the model - very open to alternative approaches here.

schedule.md
sbfnk: Yes, good idea to give them something that we got running (and show them how).

schedule.md
sbfnk: Good idea.

schedule.md
seabbs: Yeah I'm really not sure. I quite like the idea of using the play as a jumping off point for eval (i.e so how do we know which of you to trust giving advice to policy makers etc etc) but I think either could work. 

We could also have the play be basically what is good or bad about these forecasts you are making (which they will struggle to answer) and then present the eval as that answer.

schedule.md
seabbs: Here I am making a p oint that the coonvolution delays are not the same thing as nowcasting.

schedule.md
seabbs: > If the aim is to get them using the available tools rather than developing new ones 

So I think the aim is more to make them understand the underlying concepts vs making a model or using one. 

I agree that pushing people towards tools that exist makes sense when it makes sense to use them.

kathsherratt:
I'd be a bit tempted to start straight off the bat on Monday with the short intro to Rt estimation. Then widen out into delays in general and nowcasting. People will be familiar with Rt, and I think it can be good to start with that in order to introduce the general idea of more carefully thinking through and estimating latent/unobserved processes.
I'm not very attached to this suggestion and can see the logic of how it is now, so feel free to ignore.

seabbs:
So my main push back would be that the models for renewal equations are much more complicated to write down/read vs a simple distribution and so it wold be quite a hot start to the process. I would also push back a bit that people won't be familiar with say an incubation period.

I also like the idea of people know about delays ahead of Rt estimation because then right when you start talking about Rt you can make it clear its a latent quantity and its latent because of these delays you have just introduced. 

On the other hand if splitting Rt into an underlying latent process lecture (i.e just the renewal)  and then a "real-world" rt (w/ delays) after the delay bit I can see it working out.

kathsherratt:
Yeah that's all fair! Keep as is then :)

schedule.md
seabbs: I've made #22 and #21 and assigned myself for this

schedule.md
seabbs: Closing in favour of the near duplicate below and #21 and #22.

schedule.md
seabbs, line 11:
```suggestion
```

PR #17, opened 2024-04-30
First draft of delay distributions and biases session

### LINKED ISSUE #18: Practice session: Simulating biases in delay distributions and estimating delays without adjustment on these data
As the title make a first draft along the lines of #10

sbfnk:
initial skeleton for a delay distributions session. work in progress

Closes #18


Rmd/delay_distributions.Rmd
```diff
@@ -0,0 +1,139 @@
...
+library("fitdistrplus")
+## question: what to do about zeroes? Probably best to add 0.5 to all
+fitdistrplus::fitdist(c(na.omit(hosp_scaled$onset_to_hosp + 0.5)), distr = "lnorm")
+## pretty good - will need to tweak to highlight issue
+```
+
+some bias in estimation - how to avoid? Create discrete version and use fitdistrplus with that.
+
+```r
+hosp_scaled <- hosp_scaled |>
```
seabbs, line 122: This is better than above but isn't "right" so do we want to touch on that?

sbfnk: yes and the address in full estimation below with stan model I think

scripts/snippets/inf-hosp.r
```diff
@@ -0,0 +1,19 @@
+## first, choose random delays
+inf_hosp <- infection_times |>
+  mutate(
+    incubation_period = rgamma(n(), shape = 5, rate = 1),
+    onset_to_hosp = rlnorm(n(), meanlog = 1, sdlog = 1),
+    infection_date = as.Date(start_date + infection_time),
```
seabbs, line 6: if start_date is a date then this data won'
t have double censoring (as the input is discrete).

delay_distributions.Rmd
```diff
@@ -0,0 +1,206 @@
...
+
+## visualise the infection curve
+ggplot(infection_times, aes(x = infection_time)) +
+  geom_histogram() +
+  xlab("Infection time (in days)") +
+  ylab("Number of infections")
+```
+
+We would now like to simulate hospitalisations arising from this outbreak.
+We assume that the incubation period (or time from infection to symptom onset) is gamma-distributed with shape 5 and rate 1.
```
seabbs, line 42: This seems way to hard to me? Unless we are doing a lot of work in the lecture to teach what the delay process is

scripts/snippets/inf-hosp.r
seabbs: (this is why we used a continuous model to simulate infections in the delay dist long paper)

scripts/snippets/inf-hosp.r
```diff
@@ -0,0 +1,19 @@
...
+    onset_date = as.Date(start_date + infection_time + incubation_period),
```
seabbs, line 7: If infection time discrete? It needs to be continuous for double censoring to be present

scripts/snippets/inf-hosp.r
```diff
@@ -0,0 +1,19 @@
...
+    hosp_date =
+      as.Date(start_date + infection_time + incubation_period + onset_to_hosp)
```
seabbs, line 9: Is it going to be clear that `as.Date` acts as a flooring mechanism?

scripts/snippets/inf-hosp.r
```diff
@@ -0,0 +1,19 @@
...
+  )
+## next, set 70% of the hospitalization dates to NA because only 30∞ of cases
```
seabbs, line 11: Do you reeally need this complexity for the lesson at hand?

delay_distributions.Rmd
```diff
@@ -0,0 +1,206 @@
...
+
+As with all our estimation tasks throughout this course, we will be using stan, a fully-fledged toolkit for estimation.
+We will try to represent the model we simulated above as much as possible in this.
+One way that this could look, for examle, is:
+
+```{r delay_model}
+library("cmdstanr")
+
+delay_model <- '
+  data {
```
seabbs, line 99: An option here: https://mc-stan.org/cmdstanr/articles/r-markdown.html

delay_distributions.Rmd
```diff
@@ -0,0 +1,206 @@
...
+```
+
+This will probably cause an error - can you think of a reason why (if you don't see an error you could still think about what may cause an error, or resimulate the data that is being fitted to by re-running the commands that add columns to the infection times data set)?
+
+Can you change something that will avoid the error?
+You can look at [our example](delay_distribution_solution.html#estimating-delay-distributions) for a suggested fix, before moving on to a better solution in the next section.
+
+# Accounting for censoring
+
+A better way to account for censoring is to integrate over a window:
```
seabbs, line 144: I thought we discussed this and agree it was best to stick to the solution that is closest to best practice and mirrors the simulation?

delay_distributions.Rmd
```diff
@@ -0,0 +1,206 @@
...
+
+model$sample(
+  data = list(
+    n = nrow(na.omit(hosp_scaled)),
+    onset_to_hosp = na.omit(hosp_scaled)$onset_to_hosp
+  )
+)
+```
+
+You will see that this solution does not require any changes to the data in order to obtain an estimate.
```
seabbs, line 198: Is the lecture going to cover some of the why as this just feels pulled out of thin air?

seabbs (commented):
Just had another look. Not clear where this is at so comments may be a bit confused.

delay_distributions.Rmd
sbfnk: That's nice. It won't work if people prefer to copy & paste.

delay_distributions.Rmd
sbfnk: What would be the right level of difficulty in your view?

delay_distributions.Rmd
```diff
@@ -0,0 +1,206 @@
...
+## - `onset_to_hosp`: the time from symptom onset to hospitalisation
+## - `infection_date`: the date on which an individual became infected
+## - `onset_date`: the date on which an individual started showing symptoms
+## - `hosp_date`: the date on which an individual became hospitalised (or NA if
+##    they didn't get hospitalied)
+df <- NULL ## replace NULL with your code
+```
+
+```{r inf_hosp_solution}
+source(here::here("scripts", "snippets", "delay-distributions.r"))
```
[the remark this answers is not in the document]
sbfnk: Fixed in 46477ce

scripts/snippets/inf-hosp.r
sbfnk: it's continuous

scripts/snippets/inf-hosp.r
sbfnk: Possibly not. I'm not a big fan of this either. Do you have a suggestion for a clearer way of doing this (alternatively can add an explanation).

scripts/snippets/inf-hosp.r
sbfnk: Perhaps not - although it helps illustrate the point that doing so leads to a less sharp time distribution as in the figure that follows. That's a point I think we should try to get across. Do you have a suggestion for a better way of doing so?

delay_distributions.Rmd
sbfnk: Would be happy to have a solution with latent infection times but when I did so the stan code turned quite complex. I was probably missing a trick.

delay_distributions.Rmd
sbfnk: I think ideally this session would be followed by an explanation why, i.e. do first then see the theory behind it.

sbfnk:
> Just had another look. Not clear where this is at so comments may be a bit confused.

Thanks! Definitely work in progress but very useful to have the approach challenged at this stage.

delay_distributions.Rmd
seabbs: I see

scripts/snippets/inf-hosp.r
sbfnk: Turns out it actually doesn't, and this was news to me. We may have to use `lubridate` or one way or the other definitely something clearer.

``` r
date <- as.Date("2024-05-24")
fractional <- as.Date(date + 0.2)
fractional
#> [1] "2024-05-24"
as.POSIXct(fractional)
#> [1] "2024-05-24 04:48:00 UTC"
```

<sup>Created on 2024-05-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

delay_distributions.Rmd
sbfnk: One way to simplify this would be to not work with dates but instead numeric days (starting with 0). We could then introduce dates in #54.

scripts/snippets/inf-hosp.r
sbfnk: see https://github.com/nfidd/nfidd/pull/17#discussion_r1624199126

delay_distributions.Rmd
sbfnk: Suggestion from conversation outside this PR: error on 0-1 delays, which simplifies things

sbfnk:
Updated session to now address censoring and truncation separately.

The truncation model currently does not yield great estimates, though I haven't found an error yet.

stan/censored_delay_model.stan
```diff
@@ -0,0 +1,28 @@
...
+  for (i in 1:n) {
+    true_onset_to_hosp[i] =
+      onset_to_hosp[i] + hosp_day_time[i] - onset_day_time[i];
+  }
+}
+
+model {
+  meanlog ~ normal(0, 10);
+  sdlog ~ normal(0, 10) T[0, ];
+  onset_time ~ beta(1, 1);
```
seabbs, line 24: Using a beta prior makes this not quite a traditional censoring model (a rare case where uniform is maybe okay) but should only make a small difference.

stan/censored_delay_model.stan
```diff
@@ -0,0 +1,28 @@
...
+transformed parameters {
+  array[n] real<lower = 0> true_onset_to_hosp;
+  for (i in 1:n) {
+    true_onset_to_hosp[i] =
+      onset_to_hosp[i] + hosp_day_time[i] - onset_day_time[i];
+  }
+}
+
+model {
+  meanlog ~ normal(0, 10);
```
seabbs, line 22: These priors are very wide. I can see this throwing some issues potentially?

stan/truncated_delay_model.stan
```diff
@@ -0,0 +1,19 @@
...
+  real meanlog;
+  real<lower = 0> sdlog;
+}
+
+model {
+  meanlog ~ normal(0, 10);
+  sdlog ~ normal(0, 10) T[0, ];
+
+  for (i in 1:n) {
+    onset_to_hosp[i] ~ lognormal(meanlog, sdlog) T[0, time_since_onset[i]];
```
seabbs, line 17: Hmm is this equivalent to dividing by the CDF? I think it isn't but I would be hard pressed to give a ref to that effect. Is there are model problems thought this is where I would expect them to be coming from.

seabbs: Reading this again would certainly seem to imply that this works: https://mc-stan.org/docs/stan-users-guide/truncation-censoring.html#truncated-data.section

seabbs:
This looks really good. 

Its a shame not to cover both at once but I can see why. If time I suggest we code up a mode that shows this. 

> The truncation model currently does not yield great estimates, though I haven't found an error yet.

Seem my query as to if the automated notation is properly normalising. I don't know but perhaps its worth trying. I think merge this and promote that + prior exploration to an issue.

stan/censored_delay_model.stan
sbfnk: Making them narrower might actually solve the issue in the truncated model (as it's finding some parameter combinations that are so right-heavy that the denominator gets shrunk).

stan/truncated_delay_model.stan
sbfnk: When I tried both they seemed to give the same results.

sbfnk:
> Its a shame not to cover both at once but I can see why. If time I suggest we code up a mode that shows this.
> 
> >

Yes - also a good exercise to explore for those who rush through the session.

sbfnk:
> I think merge this and promote that + prior exploration to an issue.

Just waiting for your approval for merge.

seabbs (approved):

PR #33, opened 2024-05-17
intro delay distributions session

sbfnk:
First stab. Do we want posterior predictive checks / prediction plots? Probably good to teach some good practice but this is currently only a 15 minute session so might be a push.


sbfnk:
Actually 35 minutes. So might be able to squeeze something in.

sbfnk:
We may want to think about what to do with stan outputs - e.g. secretly hide them.

PR #39, opened 2024-05-28
R estimation session

sbfnk:
work in progress R estimation session


sessions/05-R-estimation-and-the-renewal-equation.qmd
```diff
@@ -3,3 +3,175 @@ title: "R estimation and the renewal equation"
...
+```
+
+::: {.callout-tip}
+The code in this session can be run as an interactive notebook using RStudio, or copied-and-pasted into an R session.
+It needs to be run inside the course repository so that the `here()` commands below find the stan model files.
+:::
+
+# The renewal equation as a process model for infectious diseases
+
+This session represents a shift from the models used in the previous session.
```
seabbs, line 36: I think this could be linked instead to the previous convolution session where we had to have a vague `Normal(0, 10)` model of infections. The renewal process is then justified as being closer to the "real" infection process and so  should help the model recover more realistic infections - especially in settings where any delays are also unknown/noisy... 

We could then discuss how this Rt is specifically also useful and often a target for inference in and of itself

sessions/05-R-estimation-and-the-renewal-equation.qmd
```diff
@@ -3,3 +3,175 @@ title: "R estimation and the renewal equation"
...
+    I[t + 1] <- sum(I_segment * gen_pmf) * R[t]
+  }
+  return(I[-1]) ## remove I0 from time series
+}
+```
+
+::: {.callout-note}
+## Take 10 minutes
+Try to understand the `renewal()` function above.
+Compare it to the `convolve_with_delay()` function from the session on convolutions.
```
seabbs, line 108:
```suggestion
Compare it to the `convolve_with_delay()` function from the session on convolutions.
How are they similar?
```

sessions/05-R-estimation-and-the-renewal-equation.qmd
sbfnk: good idea - have updated text

sessions/05-R-estimation-and-the-renewal-equation.qmd
```diff
@@ -3,3 +3,175 @@ title: "R estimation and the renewal equation"
...
+::: {.callout-tip}
+## Instantaneous vs. case reproduction number
+There are different definitions of the reproduction number that can be applied to the situation where it changes in time.
+As it is defined above it is also called the _instantanous_ reproduction number because any change affects all currently infectious individual instantaneously.
+Another example of a definition is the _case_ reproduction number, where changes affect individuals at the time that they are infected but then they have a constant reproduction number throughout their infectious period.
+:::
+
+# Simulating an epidemic using the renewal equation
+
+With the theory out of the way we now turn to simulating an epidemic using the renewal equation.
```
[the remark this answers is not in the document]
sbfnk: have added a box to clarify - there's quite a lot of reading/theory there now, which we may want to review (but also good to be clear on terminology etc.

seabbs (approved):
This is really nice and clear.

PR #79, opened 2024-06-04
render on PRs

sbfnk:
At the moment pages only get rendered upon push to main which is a bit late - with this PR that happens in PRs (without deployment) which should help catch errors.


sbfnk:
solving issues with latest merged session along the way

sbfnk:
Minor change - bypassing review.

PR #84, opened 2024-06-04
Add remaining skeleton

seabbs:
Adds remaining template text and aims to include info from the session overview into all the individual sessions

### DIFF
--- a/sessions/00-introduction-and-course-overview.qmd
+++ b/sessions/00-introduction-and-course-overview.qmd
@@ -3,3 +3,6 @@ title: "Introduction and course overview"
 order: 0
 ---
 
+# Introduction to the course and the instructors
+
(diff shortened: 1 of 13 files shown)


sbfnk (commented):
Thanks! I wouldn't have thought we need "wrap up" headings - what would be there? I don't think we need any slides/text for these but just use the time to go through the session doc.

For the headers that relate to material on slides, would the idea be that there is only a link to the slides?

seabbs:
> For the headers that relate to material on slides, would the idea be that there is only a link to the slides?

Yes. I thought we wanted to make the session file the one stop shop. 

> I wouldn't have thought we need "wrap up" headings - what would be there?

Esentially as above - if we have any summary slides we I thought we might (like 3).

sbfnk (approved):

PR #87, opened 2024-06-07
nowcasting: introduction / motivation

sbfnk:


sbfnk:
It's making a start at https://github.com/nfidd/nfidd/issues/48 but may well make sense for you to have a stab at both - in which case feel free to ditch or peruse the text here.

seabbs:
@sfnk this is a bit mad and long but I think ready for a review. As it is mad and long I am tempted to wave it through a bit more than usual moving todos to issues. 

Something to note is that when testing this I was using simulated data with Rt > 1 (i.e nowcasting into a growing outbreak) but somewhere along the line this has updated so that the nowcast is at turnover. This slightly undermines the point I was trying to make as now a trend carried forecast nowcast actually looks the worst... 

Something else to flag here is I have done a poor job expressing that the simple nowcast effectively cheats by using the known delay.

Finally, the joint nowcasts aren't really doing so hot. I think this is because they aren't doing a great job of capturing the reporting delay because the data is so sparse but I haven't investigated.

sbfnk:
As a very general point can I suggest to do one sentence per line rather than one paragraph per line? Helps `diff`/GitHub web formatting when suggesting changes.

No need to make changes here/now (can go in a separate issue for future work) but if we agree this makes sense could aim to follow when writing future sessions.

sessions/nowcasting.qmd
```diff
@@ -31,6 +32,555 @@ The code in this session can be run as an interactive notebook using RStudio, or
...
+
+We can now construct some simulated data and use this delay distribution to nowcast the most recent data. Here we use the same simulation approach as in the [renewal session](R-estimation-and-the-renewal-equation) and apply the `reporting_delay` to the last 15 days of data.
+
+```{r, load-simulated-onset}
+source(here::here("snippets", "simulate-onsets.r"))
+reported_onset_df <- onset_df |>
+  filter(day < 40) |>
+  mutate(proportion_reported = c(rep(1, n() - 15), rev(proportion_reported)),
+         reported_onsets = map_dbl(onsets * proportion_reported, \(x) rpois(1, x)) 
+  )
```
sbfnk, line 163: here I think it would be good to do this inside an R function `condition_onsets_by_report()` that does exactly the same as the stan function below.

snippets/simulate-onsets.r
```diff
@@ -0,0 +1,16 @@
...
+source(here::here("functions", "censored-delay-pmf.r"))
+source(here::here("functions", "convolve-with-delay.r"))
+
+gen_time_pmf <- censored_delay_pmf(rgamma, max = 14, shape = 4, rate = 1)
+gen_time_pmf <- gen_time_pmf[-1] ## remove first element
+gen_time_pmf <- gen_time_pmf / sum(gen_time_pmf) ## renormalise
+
+ip_pmf <- censored_delay_pmf(rgamma, max = 14, shape = 5, rate = 1)
+onsets <- convolve_with_delay(inf_ts$infections, ip_pmf)
+onsets <- purrr::map_dbl(onsets, \(x) rpois(1, x))
```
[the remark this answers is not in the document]
sbfnk: Not quite clear to me why we sample here instead of just using the delayed onsets later. I can see why we wouldn't model the expectation of the individual level process but earlier we motivated that as uncertainty from the count process - do we need to provide a different motivation here?

sessions/nowcasting.qmd
```diff
@@ -31,6 +32,555 @@ The code in this session can be run as an interactive notebook using RStudio, or
...
+  filter(day > 1) |>
+  left_join(nowcast_onsets, by = "day") |>
+  ggplot(aes(x = day, y = onsets)) +
+  geom_col() +
+  geom_ribbon(aes(ymin = lower, ymax = upper), alpha = 0.5) +
+  geom_line(aes(y = median))
+```
+
+::: {.callout-tip}
+As we found in the [using delay distributions to model the data generating process of an epidemic session](using-delay-distributions-to-model-the-data-generating-process-of-an-epidemic#estimating-a-time-series-of-infections), this simple model struggles to recreate the true number of onsets. This is because it does not capture the generative process of the data (i.e. the transmission process and delays from infection to onset). In the next section we will see how we can use a model that does capture this generative process to improve our nowcasts.
```
sbfnk, line 238: it seems to work fairly well? This may need more motivation I think.

sbfnk (commented):
Got to the start of the random walk section - posting comments on the earlier bits.

sessions/nowcasting.qmd
sbfnk: Potentially it could also be a "going further" bit, i.e. if you've got it all try putting in the geometric random walk from before which should work better and see if it did.

sbfnk:
There are a some stray `.rmarkdown` and `.html` files - add to `.gitignore`?

Also turned some 2 mins to 5 mins looking at stuff just so that people can take a breath / aren't rushed.

snippets/simulate-onsets.r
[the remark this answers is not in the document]
seabbs: Observation error? I think in general it makes sense to use a model that is a close the one we are using for fitting as possible

stan/estimate-inf-and-r-rw.stan
```diff
@@ -33,7 +33,7 @@ transformed parameters {
 
 model {
   // priors
-  init_R ~ lognormal(-.1, 0.5); // Approximatly Normal(1, 0.5)
+  init_R ~ normal(-.1, 0.5); // Approximately Normal(1, 0.5)
```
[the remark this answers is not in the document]
seabbs: I don't think so as the whole random walk gets exponentiated

snippets/simulate-onsets.r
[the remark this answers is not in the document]
seabbs: also wow who knew it took a vector (well obviously you)

sessions/nowcasting.qmd
```diff
@@ -31,6 +32,555 @@ The code in this session can be run as an interactive notebook using RStudio, or
...
+::: {.callout-note collapse="true"}
+## The solution
+- The `proportion_reported` is the cumulative distribution function of the delay distribution. It gives the probability that a report is made on day 1 or earlier, day 2 or earlier, etc. Note that for days more that 15 days into the past 
+- The `reported_onsets` are the number of onsets that are reported on each day. This is calculated by multiplying the number of onsets by the proportion of onsets that are reported on each day. It has Poisson noise added to it to simulate the stochasticity in the reporting process.
+:::
+
+We can now fit our first nowcasting model. Here we assume exactly the same generative process as we used for simulation and model the number of onsets as independent draws from a normal distribution.
+
+```{r stan-simple-nowcast}
+mod <- cmdstan_model(here("stan", "simple-nowcast.stan"), include_paths = here("stan"))
```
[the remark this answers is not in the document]
seabbs: yeah this was left over from some frantic debugging on the path issue

seabbs:
> There are a some stray .rmarkdown and .html files - add to .gitignore?

Woops. Yes good idea

stan/estimate-inf-and-r-rw.stan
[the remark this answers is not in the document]
sbfnk: Ah got it

sessions/nowcasting.qmd
seabbs: This was my point about all the text being true and then the data "changed" on a rerender and we had a decreasing outbreak -> model that works quite well. 

When you have a growing outbreak it should revert to the prior towards the end and hence look bad. 

The other related problem is that we know the delay distribution so the performance should be the best it can be

sessions/nowcasting.qmd
seabbs: Part of #122?

snippets/simulate-onsets.r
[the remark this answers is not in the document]
sbfnk: > Observation error?

but we observe the delayed/truncated reports (and that's where we have a poisson in the fitting model) - what's the rationale for adding observation error twice (once at onset and then after delay)?

sessions/nowcasting.qmd
sbfnk: might be worth adding a data point (as earlier)
image

sessions/nowcasting.qmd
```diff
@@ -31,6 +31,555 @@ The code in this session can be run as an interactive notebook using RStudio, or
...
+```
+
+The `plnorm()` function is related to the `rlnorm()` function we used earlier to simulate the individual level reporting delay, but instead it gives the cumulative distribution function rather than random samples. That is, it gives us the probability that a report is made on day 1 or earlier, day 2 or earlier, etc.
+
+We can now construct some simulated data and use this delay distribution to nowcast the most recent data. Here we use the same simulation approach as in the [renewal session](R-estimation-and-the-renewal-equation) and apply the `reporting_delay` to the last 15 days of data.
+
+```{r, load-simulated-onset}
+source(here::here("snippets", "simulate-onsets.r"))
+reported_onset_df <- onset_df |>
+  filter(day < 40) |>
```
sbfnk, line 159:
```suggestion
  filter(day < cutoff) |>
```

sessions/nowcasting.qmd
```diff
@@ -31,6 +31,555 @@ The code in this session can be run as an interactive notebook using RStudio, or
...
+source(here::here("functions", "censored-delay-pmf.r"))
+reporting_delay_pmf <- censored_delay_pmf(rlnorm, max = 15, meanlog = 1, sdlog = 0.5)
+plot(reporting_delay_pmf)
+```
+
+We can then simulate the reporting triangle:
+
+```{r simulate-reporting-triangle}
+reporting_triangle <- onset_df |>
+  filter(day < 40) |>
```
sbfnk, line 353:
```suggestion
  filter(day < cutoff) |>
```

sbfnk (commented):
can reuse the `cutoff` variable from earlier (41 instead of 40) which adds more growth

sessions/nowcasting.qmd
seabbs: > might be worth adding a data point (as earlier)

?

sbfnk:
made few commits so it compiles again (needed because #128 is rebased to this now)

sbfnk:
Merging this for now to facilitate other development based on this - further modifications can be made in new issues/PRs.

snippets/simulate-onsets.r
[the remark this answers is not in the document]
sbfnk: @seabbs still not clear why we're sampling here btw if it's not the observation (as in the nowcasting session where we truncate and then add another observation error).

PR #88, opened 2024-06-07
stan intro

sbfnk:
This revives some of the early code from https://github.com/nfidd/nfidd/pull/10 for a gentle introduction to stan


sbfnk:
Quite a lot of overlap with the following session (which I'd forgotten about) so this will require some streamlining.

PR #89, opened 2024-06-07
installing cmdstan step

sbfnk:
Adds a section on install `cmdstan` to the "Getting Setup" vignette - for people to do in advance of the course.


seabbs (commented):
This looks good to me. Do we want to recommend using check toolchain (with fix = TRUE)

PR #96, opened 2024-06-11
remove locking into session order

### LINKED ISSUE #93: rename session files
Once all the sessions are written we might want to change the names of the session files to remove the numbering - nice to see things in order while developing but will make it a pain to change/add/remove things later

sbfnk:
Closes #93 

- removes all numbering from filenames
- updates relative links
- updates references like "previous session" to instead refer to the session explicitly

This should make it easier to restructure later, should we or anyone else want to do so.


seabbs (approved):
LGTM

PR #113, opened 2024-06-17
Fixes getting setup

seabbs:
I decided to switch to working with Rstudio this week to get the participant experience. Did the getting set up workflow and made some corrections. 

Made the choice to switch to `pak` as may save some issues with people getting setup.

### DIFF
--- a/getting-set-up.qmd
+++ b/getting-set-up.qmd
@@ -4,17 +4,17 @@
 
 # Installation of the `nfiidd` package
 
-To install the packages needed in the course, including the `nfiidd` package that contains data files used, you can use the `remotes` package:
+To install the packages needed in the course, including the `nfiidd` package that contains data files used, you can use the `pak` package:
 
 ```{r install, eval=FALSE}
-install.packages("remotes")
-remotes::install_github("nfiidd/nfiidd", dependencies = TRUE)
+install.packages("pak")
+pak::pak("nfidd/nfidd")
 ```
 
 Then you can check that the installation completed successfully by loading the package into your **R** session:
 
 ```{r load, eval=FALSE}
-library("nfiidd")
+library("nfidd")
 ```
 
 # Installing `cmdstan`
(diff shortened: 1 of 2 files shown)


seabbs:
I'm now trying to run the sessions and finding that none of the stan models will compile due to not being able to find the include functions

getting-set-up.qmd
```diff
@@ -23,7 +23,7 @@ The course relies on running stan through the `cmdstanr` **R** package, which it
 This requires a separate installation step:
 
 ```{r cmdstan_install, eval = FALSE}
-cmdstanr::install_cmdstan()
+cmdstanr::install_cmdstan(cores = 2)
```
sbfnk, line 26: why 2?

seabbs: so that is compiles faster and under the assumption they will have at least two we can use?

seabbs:
> Given that it works in the CI, is this an RStudio specific issue?

No idea!

seabbs:
Bug is unrelated to the course. It looks like the issue with spaces in file paths has come up again (I was working out of a google drive folder called My Drive - moving the folder fixes it.

sbfnk:
That's really good to know (especially as it may well affect course participants).

I wonder if it's something we could raise/fix upstream at some point.

seabbs:
So what is super weird is they have fixed it (i.e there is a closed issue from a year ago) so it is super weird it has really appeared with all new installs of everything.

getting-set-up.qmd
sbfnk: minor point of course but I'm not sure why we'd overrule the `cmdstanr` default here. Also, every extra argument adds complexity so I think we should only use what's definitely necessary (i.e. not this).

sbfnk, line 26:
```suggestion
cmdstanr::install_cmdstan()
```

seabbs: the stan default is 1 core and so slow is the reason. People new to cmdstan won't know it exists. I don't really mind either way

seabbs: I don't really understand why we would want people to have to wait longer when installing but I also don't really mind

sbfnk (approved):

PR #115, opened 2024-06-17
No rounding

sbfnk:
rounding makes the likelihood non-smooth which causes fitting issues as referenced in https://github.com/nfidd/nfidd/pull/112


seabbs (approved):
Nice. Panic over.

PR #132, opened 2024-06-19
simplify random walk model in convolution session

### LINKED ISSUE #127: Forecasting in the renewal eqn session
I think my view is that we shouldn't include the forecasting piece in the random walk model in the renewal equation session (introduced in #112). 

For me it interrupts the nice flow of building up the model and comparing line by line - especially by adding the `transformed data` and `generated quantities` blocks which aren't introduced at this point, unlike other blocks previously. 

Not the biggest deal but in an already busy session it kinda just raises questions that aren't answered. So I don't quite see the advantage, compared to adding forecasting into the stan code at the time that it's first used (I assume in [forecasting](https://github.com/nfidd/nfidd/pull/126/files#r1645219410)).

I know it means adding another stan model to the lineup but this doesn't really affect the learner. Is this a big barrier? I'm open to understanding more the motivation for introducing it earlier.

### LINKED ISSUE #131: random walk model for R estimation struggles to fit
The estimates are great but the MCMC struggles, sometimes destructively so. Investigating whether any reparametrisation can solve it (quite a bit of correlation between parameters to wreak havoc with HMC).

sbfnk:
Closes #131

I think the cumulative sum creates struggles for the sampler (and experimentation confirms this). Modelling the random walk on R directly removes divergent transitions and simplifies the model code.

If we agree to go ahead with this I'll update the text in the session accordingly.

Closes #127


stan/estimate-inf-and-r-rw.stan
```diff
@@ -12,38 +11,24 @@ data {
...
-  obs ~ poisson(onsets[1:n]);
-}
-
-generated quantities {
-  array[h] real forecast;
-  if (h > 0) {
-    for (i in 1:h) {
-      forecast[i] = poisson_rng(onsets[n + i]);
-    }
+  logR[1] ~ normal(0, 0.5);
```
seabbs, line 28: It seems surprising that this works better given that it is a centred parameterization. Any evidence? If we are going with this then need to update the text as it talks about a non-centred random walk.

seabbs: I think if we do want to do this then we need to wrap it in a function with the same rational that everything else is in a function

sbfnk: > It seems surprising that this works better given that it is a centred parameterization. It could be because there is so much data supporting the rw? Any evidence?

It might need more investigation but the thing that I suspect might cause problems in the existing parametrisation is that a change in `rw_sd` directly scales all the R values (which cumulatively and exponentially enters the likelihood) whereas here it only affects the priors on R. The only evidence is that the previous parametrisation would cause divergent transitions and trajectories going haywire whereas this one doesn't (consistently across a few repetitions).

sbfnk: > I think if we do want to do this then we need to wrap it in a function with the same rational that everything else is in a function

I think the main rationale for having things in functions is that we can have corresponding R functions doing the same things which helps us explain what happens. Probably indeed a good idea to do this here too, i.e. simulate a random walk first with an R function and then use it here, only that it's perhaps more difficult to motivate as contrary to e.g. the delays we know reproduction numbers aren't doing a random walk, it's just a better model than independent draws.

sbfnk: the parametrisation here has no divergent transitions but low ebfmi which is not ideal - but more importantly it doesn't produce divergent trajectories which the existing one frequently does.

These are the means from 20 fits ("direct" is the model proposed here):
```r
# A tibble: 2 × 4
  approach num_divergent num_max_treedepth ebfmi
  <chr>            <dbl>             <dbl> <dbl>
1 direct          0.0125               0   0.188
2 original       10.5                 45.9 0.978
```

seabbs: I struggle to believe this based on passed models - I don't think I've ever seen this to be the case with a random walk. 

> suspect might cause problems in the existing parametrisation is that a change in rw_sd directly scales all the R values (which cumulatively and exponentially enter the likelihood) whereas with the change it only affects the priors on R. 

I don't think this shakes out as being that different? 

> I don't think this is really centred) - if not I think we should stick with this as it's easier to read.

I'm not sure I follow but what is implemented in `main` is non-centred what is here is centred and your suggestion (that is crossed out) is something else I think. 

> I think the main rationale for having things in functions is that we can have corresponding R functions doing the same things which helps us explain what happens.

I don't really agree with this? The main reason I see to have functions is to modularise the code/make it easier to understand.

> but more importantly it doesn't produce divergent trajectories which the existing one frequently do

Could you clarify what you mean here by divergent trajectories? Are they trajectories that are outside the support of the prior or just large trajectories?

> ebfmi

I think this being so low is a problem but I agree the difference in dts in quite convincing. In general, I would be surprised if we can get away with not cranking up adapt delta for autoregressive models

sbfnk: > I'm not sure I follow but what is implemented in main is non-centred what is here is centred and your suggestion (that is crossed out) is something else I think.

yes agree

sbfnk: > I don't really agree with this? The main reason I see to have functions is to modularise the code/make it easier to understand.

Yes, but then I'm not convinced that 4 lines of code for a random walk is that hard to understand. Don't feel that strongly though.

sbfnk: > Could you clarify what you mean here by divergent trajectories? Are they trajectories that are outside the support of the prior or just large trajectories?

Trajectories that diverge into nonsensical spaces, e.g. millions of infections.

sbfnk: The other thing to mention is that the version here is much faster. Perhaps it's because it explores less of the posterior (see ebfmi). But it doesn't produce wild trajectories either (in 20 iterations) whereas the existing version does in many cases.

seabbs: > Trajectories that diverge into nonsensical spaces, e.g. millions of infections.

But this doesn't mean its an "error" it could just be that the prior model is stupid (which it is)

seabbs: > The other thing to mention is that the version here is much faster. Perhaps it's because it explores less of the posterior (see ebfmi). But it doesn't produce wild trajectories either (in 20 iterations) whereas the existing version does in many cases.

I think we need to try and split differences here between diagnostic issues and model misspecification issues (i.e a random walk is a bad more here as we know Rt is just going down and not brownian motion)

sbfnk: Agree but the model is the same in all cases, only the parameterisation different - and one "works" better than the other in terms of the sampler being happier and faster and outcomes looking good, unless we think the low ebfmi is such an issue that we don't want to go there. This may be the case but one way or the other we should either shield participants from getting nonsensical estimates (which this does) or explain what to do when they arise.

sbfnk: F2F: we'll definitely change something (at the minimum the cumulative sum to a for loop) - to be revisited at a later stage in terms of what exactly works best.

sbfnk:
re diverging trajectories these are some examples (on current render of the website):

image

image

seabbs:
yup that is not a happy bunny. Happy to change the formulation here. Or we can tighten the initial prior as that should restrict this?

sbfnk:
> Or we can tighten the initial prior as that should restrict this?

on `rw_sd`? Or we could fix it - solves all problems but really hard to motivate what to use if using for forecasts.

seabbs:
> Or we could fix it -

I really don't want to do that! 

No I mean the prior for the initial R value as it seems like this must be correlated for these weird trajectories to happen?

PR #157, opened 2024-06-20
Forecasting concepts - suggestions

kathsherratt:
mainly adding titles to plots & typos

just seen part overlap with #150 but doesn't conflict.


seabbs:
Due to the use of squash commits this will have a very messy merge life. I'd suggest waiting until we merge in forecasting concepts and do as a supplement?

kathsherratt:
Yes agree, just making suggestions while I was working through it - we'll just come back to this when the session is more final.

seabbs:
I think we can wake this back up now? Sorry lots of merge issues and stuff

kathsherratt:
No problem at all I'll have another look shortly

seabbs (approved):
Very nice

Issue #122, opened 2024-06-18
ensure we have an R function for every stan function


sbfnk:
initial feedback indicates that this is very useful

seabbs:
I'm going to lightly pushback on this that in some instances this is quite a lot of work and not really key to teaching the understanding we are going for. For example exactly tracking how the joint nowcast model works isn't critical to learn the lessons we are trying to impart here?

sbfnk:
Yes, that makes sense - will bear in mind when reviewing the nowcasting session in detail but assume for now we only want this for functions we're actually using for simulation.

sbfnk:
I'll have a look at adding this for the random walk

seabbs:
I meant to add a section for the random walk like we have for the AR in the forecasting concepts but likely forgot

PR #163, opened 2024-06-20
Forecasting models

### LINKED ISSUE #52: Practice session: Forecasting with a mechanistic model
To be decided what the range is here.

### LINKED ISSUE #53: Practice session: Evaluating forecasts from a range of common models

seabbs:
This PR contains the contents for the forecasting models section but I have yet to generrate the forecasts accorss multiple dates so these are not present.

Closes #52 
Closes #53


sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
 If not using the Visual Editor please remember that the code in the session needs to be run inside the course repository so that the `here()` commands below find the stan model files.
 :::
 
+# An overview of forecasting models
+
+See the slides for an overview of forecasting models.
```
sbfnk, line 36: add paragraph in new issue (I volunteer)

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+
+# What did we learn about the random walk model from the [forecasting concepts session](forecasting-concepts)?
+
+- It was unbiased when the reproduction number was constant but when the reproduction number was reducing due to susceptible depletion, the random walk model overestimated the number of cases systematically.
+- It did relatively well at short-term forecasting indicating that it was capturing some of the underlying dynamics of the data generating process.
+
+::: {.callout-note, collapse = TRUE}
+## What did a geometric random walk model look like again?
```
sbfnk, line 44: to write function where geometric random walk is first introduced #122

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+However, in many forecasting scenarios, we may not have a good understanding of the underlying mechanisms, or the data may be too noisy to make accurate predictions.
+The worst case is that we have mistaken beliefs about the underlying mechanisms and use these to make predictions which are systematically wrong and misleading.
+In these cases, forecasters of often use _statistical models_ to make predictions which have little or no mechanistic basis.
+In our work, we have found that a combination of mechanistic and statistical models can be very powerful but that identifying the best model for a given forecasting task can be challenging.
+
+## Adding more mechanistic structure to the renewal model
+
+One way to improve the renewal model is to add more mechanistic structure. In the [forecasting concepts session](forecasting-concepts), we saw that the renewal model was unbiased when the reproduction number was constant but that it overestimated the number of cases when the reproduction number was reducing due to susceptible depletion.
```
sbfnk, line 76:
```suggestion
One way to improve the renewal model is to add more mechanistic structure. In the [forecasting concepts session](forecasting-concepts), we saw that the renewal model was making unbiased forecasts when the reproduction number was constant but that it overestimated the number of cases when the reproduction number was reducing due to susceptible depletion.
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+  geom_line() +
+  labs(title = "Simulated data from an SIR model",
+       x = "Time",
+       y = "R")
+```
+
+The key assumptions we are making here are: 
+
+- The population is constant and we rouchgly know the size of the population.
+- The reproduction number is constant (aside from the effect of susceptible depletion)
```
sbfnk, line 118:
```suggestion
- The reproduction number only changes due to susceptible depletion
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+- The number of cases is proportional to the number of susceptibles in the population.
```
sbfnk, line 119:
```suggestion
- The number of new cases at each time is proportional to the number of susceptibles in the population.
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+
+We've coded this up as a stan model in `stan/mechanistic-r.stan`. See `stan/functions/pop_bounded_renewal.stan` for the function which calculates the reproduction number. Lets load the model:
+
+```{r}
+mech_mod <- cmdstan_model(here("stan/mechanistic-r.stan"))
```
sbfnk, line 124:
```suggestion
mech_mod <- cmdstan_model(here("stan", "mechanistic-r.stan"))
```
does your version work on windows (which uses backslash)?

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+mech_mod$print(line_numbers = TRUE)
+```
+
+## Adding more statistical structure to the renewal model
+
+Rather than adding more mechanistic structure to the renewal model, we can add more statistical structure.
```
sbfnk, line 130:
```suggestion
Adding more mechanistic structure is not always possible and, if we don't specify mechanisms correctly, might make forecasts worse.
Rather than adding more mechanistic structure to the renewal model, we could add more statistical structure.
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+Before we do this, we need to think about what we want from a forecasting model.
+As we identified above, we want a model which is unbiased and which has good short-term forecasting properties.
+We know that we want it to be able to adapt to trends in the reproduction number and that we want it to be able to capture the noise in the data.
+A statistical term that can be used to describe capturing a trend is saying that the time series is _non-stationary_.
+
+The random walk model we used in the [forecasting concept](forecasting-concepts) session is a special case of a more general class of models called _autoregressive (AR) models_. 
+AR models are a class of models which predict the next value in a time series as a linear combination of the previous values in the time series.
+The random walk model is specfically a special case of an AR(1) model where the next value in the time series is predicted as the previous value, multipled by a value between 1 and -1 , plus some noise. For the log-transformed reproduction number ($ log(R_t) $), the model is:
```
sbfnk, line 138:
```suggestion
The random walk model is specifically a special case of an AR(1) model where the next value in the time series is predicted as the previous value, multiplied by a value between 1 and -1 , plus some noise. For the log-transformed reproduction number ($ log(R_t) $), the model is:
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+
+ggplot(data, aes(x = t, y = R)) +
+  geom_line() +
+  labs(title = "Simulated data from an exponentiated AR(1) model",
+       x = "Time",
+       y = "R")
+```
+:::
+
+However, we probably don't want a model which is biased towards a reproduction number of 1 (though as we know the future in this case it actually could be a reasonable choice!). So what should we do?
```
sbfnk, line 169:
```suggestion
However, we probably don't want a model which is biased towards a reproduction number of 1 (unless we have good reason to believe that is the expected behaviour). So what should we do?
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
 If not using the Visual Editor please remember that the code in the session needs to be run inside the course repository so that the `here()` commands below find the stan model files.
 :::
 
+# An overview of forecasting models
+
+See the slides for an overview of forecasting models.
+
+# What did we learn about the random walk model from the [forecasting concepts session](forecasting-concepts)?
+
+- It was unbiased when the reproduction number was constant but when the reproduction number was reducing due to susceptible depletion, the random walk model overestimated the number of cases systematically.
```
sbfnk, line 40: was it really ever constant? I would have thought than on the log scale there is a similar amount of overestimation anywhere given more or less linear decline of R.

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+mech_mod$print(line_numbers = TRUE)
+```
+
+## Adding more statistical structure to the renewal model
+
+Rather than adding more mechanistic structure to the renewal model, we can add more statistical structure.
+Before we do this, we need to think about what we want from a forecasting model.
+As we identified above, we want a model which is unbiased and which has good short-term forecasting properties.
+We know that we want it to be able to adapt to trends in the reproduction number and that we want it to be able to capture the noise in the data.
+A statistical term that can be used to describe capturing a trend is saying that the time series is _non-stationary_.
```
sbfnk, line 134:
```suggestion
A statistical term that can be used to describe capturing a trend is saying that the time series is _non-stationary_.
More specifically, a _stationary_ time series is defined as one that looks indistinguishable at all times.
In infectious disease epidemiology, this would only be expected for endemic diseases without external seasonal influence.
```

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+  geom_line() +
+  labs(title = "Simulated data from an exponentiated AR(1) model",
+       x = "Time",
+       y = "R")
+```
+:::
+
+However, we probably don't want a model which is biased towards a reproduction number of 1 (though as we know the future in this case it actually could be a reasonable choice!). So what should we do?
+
+Returning to the idea that the reproduction number is a _non-stationary_ time series, we can use a method from the field of time series analysis called _differencing_ to make the time series stationary. This involves taking the difference between consecutive values in the time series. For the log-transformed reproduction number, this would be:
```
sbfnk, line 171: needs motivation for why we want to make it stationary?

sessions/forecasting-models.qmd
```diff
@@ -31,6 +31,631 @@ The best way to interact with the material is via the [Visual Editor](https://do
...
+Some important things to note about these forecasts:
+
+  - We used a 14 day forecast horizon.
+  - Each forecast used all the data up to the forecast date.
+  - We generated 1000 posterior samples for each forecast.
+  - We started forecasting 3 weeks into the outbreak and then forecast every 7 days until the end of  the data (excluding the last 14 days to allow a full forecast).
+  - We used pregenerated simulated data:
+
+```{r}
+example_onset_df
```
sbfnk, line 369: is this different from `onset_df` above?

sbfnk (approved):
Really nice! Will take a more detailed look post-merge, here some initial comments.

sbfnk:
Closes #52 
Closes #53

seabbs:
FYI I am waiting on model fits then will update

seabbs:
I've just updated this with the fits and solutions. 

In general I am quite surprised by the difference AR model. The idea is that it should react to changes in trend (i.e the turnover) faster as it can extrapolate from small changes. This does not currently happen. I would like to tweak this as I think its a useful learning point but I also think its fine as is. 

As this has already been reviewed I will merge once all rendering successfully but have made to new issues to 1. review the AR performance and 2. review the content of this complete session.

PR #164, opened 2024-06-20
introduction to stan

sbfnk:
slides

### DIFF
--- a/sessions/slides/introduction-to-stan.qmd
+++ b/sessions/slides/introduction-to-stan.qmd
@@ -1,0 +1,105 @@
+---
+title: "Introduction to stan"
+author: "Nowcasting and forecasting of infectious disease dynamics"
+format:
+  revealjs:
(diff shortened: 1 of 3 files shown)


sessions/slides/introduction-to-stan.qmd
```diff
@@ -0,0 +1,105 @@
...
+```r
+set.seed(12345)
+```
+
+## What is stan and why do we use it
+
+- a *Probabilistic Programming Language* for Bayesian inference
+  (i.e., a way to write down models)
+  
+- models are written in a text file (often ending `.stan`) and then loaded into an R/python/etc interface
```
seabbs, line 19:
```suggestion
- models are written in a text file (ending `.stan`) and then loaded into an R/python/etc interface
```
might just be confusing

seabbs (approved):
Looks good for a first pass.

PR #179, opened 2024-06-21
Sessions update

sbfnk:
also split nowcasting session in 2.


seabbs (approved):
This is great. It looks like the issue is missing `onset_df` in the joint nowcasting session. I suggest we side step this by starting to use the saved `example_onset_df` from the forecasting models section.

seabbs:
> Not necessarily opposed but why use a saved object vs re-generating it building on the simpler versions?

Just because it means we are repeatedly regenerating it but if the seed is fixed you are right it shouldn't make a difference

sbfnk:
In the earlier sessions the simulation is part of the walkthrough so if changing to loading we'd have to change those sessions. 

> make sure it doesn't change when reloaded

are we happy for it to change between sessions (I would have thought yes but if we think it's an issue could re-set the seed inside the script)?

PR #196, opened 2024-06-21
Suggests to depends

### LINKED ISSUE #135: Make suggests depends
We need packages like `cmdstanr` to be installed - moving from Suggest to Depends would fix that.

seabbs:
Closes #135 and updates the author list. Removed the `cph` as I don't think its accurate?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Authors@R: c(
     person(
       given = "James Mba",
       family = "Azam",
-      role = "ctb",
+      role = "aut",
       email = "james.azam@lshtm.ac.uk",
       comment = c(ORCID = "0000-0001-5782-7330")
     )
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -4,31 +4,10 @@ Title: Material to support course on nowcasting and forecasting of infectious
 Version: 0.0.0.9999
 Authors@R: c(
     person(
-      given = "Sebastian",
-      family = "Funk",
+      given = "NFIDD course contributors",
+      family = "",
       role = c("cre", "aut"),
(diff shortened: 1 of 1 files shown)


sbfnk:
> Removed the cph as I don't think its accurate?

It isn't - tanks for spotting.

That said, maintaining this could become annoying (what is the relationship of the R package to the course content? has everybody who has contributed to the course contributed to the package? etc). Perhaps we should just have a generic "NFIDD course contributors" author or the like? Counterpoint being that it might be useful to link the ORCID?

sbfnk:
Given that we're only using this for dependency management and data loading it really is an abuse of R package infrastructure and probably could be solved more elegantly at a later stage.

seabbs:
> "NFIDD course contributors" author or the like? Counterpoint being that it might be useful to link the ORCID?

Yeah I think this is the right call

sbfnk (approved):

DESCRIPTION
```diff
@@ -4,11 +4,10 @@ Title: Material to support course on nowcasting and forecasting of infectious
...
     person(
-      given = "Sebastian",
-      family = "Funk",
-      role = c("cre", "cph", "aut"),
-      email = "sebastian.funk@lshtm.ac.uk",
-      comment = c(ORCID = "0000-0002-2842-3406")
+      given = "NFIDD course contributors",
+      family = "",
+      role = c("cre", "aut"),
+      email = "epiforecasts@gmail.com"
```
sbfnk, line 10: do we need an email? I don't think anyone monitors that.

sbfnk: just checked - 5 emails since 2022. Phew, nothing missed.

seabbs: I wasn't clear you could avoid having an email?

PR #198, opened 2024-06-21
Review of session on delay distributions

jamesmbaazam:
This PR:
- fixes some sentence structures,
- formats a section with line breaks,
- fixes an instance of partial matching,
- Improves a plot by removing the negative x-axis breaks, and
- makes a code block more readable by removing repeated code.

### DIFF
--- a/sessions/delay-distributions.qmd
+++ b/sessions/delay-distributions.qmd
@@ -153,8 +153,11 @@ Do you recover the parameters used for simulation?
 ::: {.callout-note collapse="true"}
 ### Hints
 
-Tip 1: First create a new column giving the time from onset to hospitalisation Tip 2: As in the last session, we can use our dataset to sample from the model posterior.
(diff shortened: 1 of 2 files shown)


sessions/delay-distributions.qmd
```diff
@@ -42,15 +42,15 @@ The outbreak lasted around 90 days after the first infection, with new infection
 
 ::: callout-note
 For now we will not concern ourselves with the model used to generate the epidemic.
-This represents a typical situation in the real world, where we may have a model of how an infection has spread, but we don't know necessary know how well this corresponds to what really happened.
+This represents a typical situation in the real world, where we may have a model of how an infection has spread, but we don't necessary know how well this corresponds to what really happened.
```
seabbs, line 45:
```suggestion
This represents a typical situation in the real world, where we may have a model of how an infection has spread, but we don't necessarily know how well this corresponds to what really happened.
```

seabbs (approved):
Nice looks good. One suggestion to make a sentence make sense.

Issue #165, opened 2024-06-20
remove coding exercises


sbfnk:
We only have a few of these in there anyway but I don't think this works with the Rstudio editable notebook set up - it requires people to do things in their own separate R session.

In the model fitting course part of the philosophy is that people should e.g. code their own Metropolis-Hastings algorithm and as a way to reinforce learning - but I have no evidence that it does so.

All that said we definitely don't need people to spend time working out how to round numbers down etc.. In summary let's just give them the code.

seabbs:
I can't see any more of these do we think this is done?

PR #199, opened 2024-06-21
Delays session - add introduction & remove coding exercises

kathsherratt:
- Adds motivating introduction - #144
- Removes coding exercises - #165

Other fixes
- Adds headers and a bit of clarifying text
- Adds comments to the `onset-hosp.R` snippet to clarify what's going on there

This crossed with @jamesmbaazam review in the meantime but attempted to merge the two. I'll take a break and come back in a bit so we don't cross again!

### DIFF
--- a/sessions/delay-distributions.qmd
+++ b/sessions/delay-distributions.qmd
@@ -12,7 +12,7 @@ set.seed(123)
 ## Objectives
 
 The aim of this session is for you to familiarise yourself with the concept of delay distributions used to describe reporting in infectious disease epidemiology.
-You will do this by working in `R` to simulate delays in the reporting of cases in an infectious disease outbreak, before using the simulation model in stan to estimate delays from a data set of outcomes.
+First we'll look at this working forwards from infections - we'll use `R` to simulate delays in the reporting of cases in an infectious disease outbreak, before using the simulation model in stan to estimate delays from a data set of outcomes.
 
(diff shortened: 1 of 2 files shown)


sessions/delay-distributions.qmd
```diff
@@ -9,10 +9,14 @@ order: 2
 set.seed(123)
 ```
 
-## Objectives
+## Introduction
+
+Every outbreak starts with an infection, but we rarely observe these directly. Instead,  there are many types of epidemiological events that we might observe that occur after the time of infection: symptom onsets, hospitalisations and discharge, etc. The length of time between any of these events might be highly variable and uncertain. However, we often need to have some estimate for these delays in order to understand what's happening now and predict what might happen in the future. We capture this variability using probability distributions, which is the focus of this session.
```
seabbs, line 14:
```suggestion
Every outbreak starts with an infection, but we rarely observe these directly. Instead, there are many types of epidemiological events that we might observe that occur after the time of infection: symptom onsets, hospitalisations and discharge, etc. The length of time between any of these events might be highly variable and uncertain. However, we often need to have some estimate for these delays in order to understand what's happening now and predict what might happen in the future. We capture this variability using probability distributions, which is the focus of this session.
```

seabbs (approved):
This looks good I think. One typo

Issue #206, opened 2024-06-22
Explore the performance of the differenced AR


seabbs:
As noted in #163 the expectation of using the difference AR was that it would do better during periods of change as it could extrapolate small changes further into time (i.e turnover faster). I think ideally we would investigate this as its a useful teaching point we are currently missing (about mechanism and non-mechanism being able to achieve the same thing)

seabbs:
One explanation could be that the prior for the standard deviation is too tight as it is the same as the random walk version but now being applied on the difference scale.

seabbs:
Still very confused about this. @sbfnk does the code look right to you/any ideas?

sbfnk:
will have a look when in a little bit when I've got slightly more time

seabbs:
I'm also planning on having another crack but I'm concerned I am missing something obvious

PR #207, opened 2024-06-22
Fixes AR formula

seabbs:
As the title. Aims to address #206. Links to #163 but I still think we should merge that asap to facilitate edits etc.


seabbs:
Just updating the forecasts using this

seabbs:
Still updating these so will ping tomorrow morning

seabbs:
I think this needs to be merged ASAP but I am quite confused about what the AR model is doing. It should be able to better track the downturn than this. The investigation I need to do is look at R in the model and see what is going on but haven't had time yet

sessions/forecasting-models.qmd
```diff
@@ -232,7 +232,7 @@ We will now use the mechanistic and statistical models to forecast the number of
 source(here::here("snippets", "simulate-onsets.r"))
 onset_df
 
-cutoff <- 71
+cutoff <- 41
```
sbfnk, line 235: intended?

seabbs: the run time was annoying me (because its 71 + 28) so I changed. I think it being quick is worth the inconsistency

stan/statistical-r.stan
```diff
@@ -37,7 +37,7 @@ model {
   init_R ~ normal(-.1, 0.5); // Approximately Normal(1, 0.5)
   rw_noise ~ std_normal();
   rw_sd ~ normal(0, 0.05) T[0,];
-  damp ~ normal(0.5, 0.25) T[0, 1];
+  damp ~ normal(0.25, 0.25) T[0, 1];
```
seabbs, line 40: maybe should have kept this higher and reduced rw_sd if needed?

seabbs: this is also quite a bad name as the higher it gets the less its damped...

sessions/forecasting-models.qmd
sbfnk, line 235:
```suggestion
# we'll make a forecast on day non day 41, pretending we haven't seen the later data
cutoff <- 41
```

sbfnk:
Only thoughts are

- does it make sense to restrict `damp` to >0? 
- this is an even steeper likelihood surface than earlier so I'd expect we'll need to crank `adapt_delta` right up 

Removing all bounds and with `adapt_delta = 0.99` I get no divergent transitions. What specifically are you unhappy with?

seabbs:
> Removing all bounds and with adapt_delta = 0.99 I get no divergent transitions. What specifically are you unhappy with?

Its not the model diagnostics its the performance of the model. We expect it to continue trends in the Rt. So if Rt is trending down it should extrapolate that. That means it should do a better job of catching changes in trend i.e. downturns and it really doesn't

seabbs:
> does it make sense to restrict damp to >0?

Yes otherwise it supports trend reversal which I think is not what we want (i.e if its has gone up in the last one it will go down in the next one). You might want that for a more general model but not here I think

seabbs:
So my though is we may need the prior on `damp` to be higher i.e `Normal(0.75, 0.25)` and maybe reduce the `rw_sd` as well?

seabbs:
As flagged I think we should merge this as it addresses the actual bug in the code (the * -> +) and then try to do an improvement in how it. works from there

sbfnk (approved):

sbfnk:
Maybe expecting a 28 step unconstrained random walk to make good forecasts is too much? It seems this could easily go wildly astray when not constrained by data and in the absence of any mean reversal.

seabbs:
> Maybe expecting a 28 step unconstrained random walk to make good forecasts is too much? It seems this could easily go wildly astray when not constrained by data and in the absence of any mean reversal.

I mean isn't that the point (that thee random walk is very bad at forecasting longer horizons).

PR #212, opened 2024-06-22
Add initial conditions for the random walk everywhere.

### LINKED ISSUE #131: random walk model for R estimation struggles to fit

seabbs:
This PR closes #131 by adding initial conditions for `init_R` and `rw_std`. I found that this was more stable than just doing so for `init_R` and didn't add any bias I could see.

I also added a note when we first do this explaining what we are doing. 

In this PR I did not update the forecasting models session as this is already covered in #207.

### DIFF
--- a/sessions/R-estimation-and-the-renewal-equation.qmd
+++ b/sessions/R-estimation-and-the-renewal-equation.qmd
@@ -349,10 +349,14 @@ data <- list(
 )
 r_rw_inf_fit <- rw_mod$sample(
   data = data, refresh = ifelse(interactive(), 50, 0), show_exceptions = FALSE, show_messages = FALSE,
-  parallel_chains = 4, max_treedepth = 12
+  parallel_chains = 4, max_treedepth = 12, init = \() list(init_R = 0, rw_sd = 0.01)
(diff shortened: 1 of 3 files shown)


sbfnk:
can we do `init = 0` instead instead of an anonymous function? See https://github.com/nfidd/nfidd/issues/131#issuecomment-2184038278

sbfnk:
> can we do `init = 0` instead instead of an anonymous function? See [#131 (comment)](https://github.com/nfidd/nfidd/issues/131#issuecomment-2184038278)

also as noted there I think we should remove the constraints (not needed with sensible init)

seabbs:
> would init=0 also work? As elsewhere I think the simpler the better here

That sets everything to zero? 

I think yes we can do that but I would rather we merge this and merge #163 first as I am almost finished updating the models there with these settings (and it takes a few hours).

sbfnk (approved):

PR #221, opened 2024-06-22
update intro sections

### LINKED ISSUE #36: Set `line_numbers = TRUE` as a central option
This is likely to prove useful in the course for communicating so should recommend setting centrally in the initial session.

### LINKED ISSUE #159: `set.seed(1234)` for the top of the code?
Minor points. 
I am always assured if prepared pages can be exactly reproduced. 

When I ran the code and displayed values were different from the ones in the web page, I would lose some degree of confidence. So, I prefer to put the `set.seed(1234)` or something like that just after reading libraries to secure the exact reproducibility.

sbfnk:
Closes #159 
Closes #36 

This required quite a bit of copying & pasting which is not ideal - does visual editor deal well with knitting chunks specified in separate files? If yes these would be a good option.

### DIFF
--- a/sessions/R-Stan-and-statistical-concepts.qmd
+++ b/sessions/R-Stan-and-statistical-concepts.qmd
@@ -93,7 +93,7 @@ To do so, we first load in the model.
 ### load gamma model from the session directory
 mod <- cmdstan_model(here("stan", "gamma.stan"))
 ### show model code with line numbers
-mod$print(line_numbers = TRUE)
+mod
 ```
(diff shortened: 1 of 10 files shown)


seabbs (approved):
Yeah that isn't ideal. I don't know. Seems like should keep as is and look at post course?

sbfnk:
> Seems like should keep as is and look at post course?

Yeah I think that applies to most non-urgent issues at this stage.

PR #223, opened 2024-06-22
Renewal session plots

kathsherratt:
Plotting fixes
- first plot currently overplots infections and R. Fix to only show R (?)
- groups plots by .draw to show trajectories
- labels plots with each model so it's easier to flip back and forth comparing
- consistently uses 100 samples in plots

Also
- removes old text/function for plotting quantiles
- renames the first two models so we don't use `mod` twice


sbfnk (approved):

sessions/R-estimation-and-the-renewal-equation.qmd
```diff
@@ -144,8 +144,8 @@ gen_time_pmf <- gen_time_pmf / sum(gen_time_pmf) ## renormalise
 As always we first load the stan model and spend some time trying to understand it.
 
 ```{r stan_estimate_r}
-mod <- cmdstan_model(here("stan", "estimate-r.stan"))
-mod
+r_mod <- cmdstan_model(here("stan", "estimate-r.stan"))
+r_mod
```
seabbs, line 148: the renaming here doesn't happening when the model is fit (line 175?). I think it needs to be the same for both?

sessions/R-estimation-and-the-renewal-equation.qmd
```diff
@@ -216,11 +218,11 @@ onsets <- convolve_with_delay(inf_ts$infections, ip_pmf)
 obs <- rpois(length(onsets), onsets)
 ```
 
-We now add this to our renewal equation model:
+We now add this to our renewal equation model and use this to estimate infections as well as the reproduction number:
 
 ```{r stan_estimate_inf_and_r}
-mod <- cmdstan_model(here("stan", "estimate-inf-and-r.stan"))
-mod
+r_inf_mod <- cmdstan_model(here("stan", "estimate-inf-and-r.stan"))
```
seabbs, line 224: same comment here the model that is being fit doesn't have the same name

sessions/R-estimation-and-the-renewal-equation.qmd
kathsherratt: yep looks like this was introduced by merging in `main`, fixing

sbfnk: sorry!

PR #224, opened 2024-06-22
introduce double interval censoring

### LINKED ISSUE #82: Consistently use "double censored"
At the moment we talk about censoring, discretisation and double interval censoring. We are really only dealing with the latter in this course and so should use this consistently as there are valid discretisation approaches that are not double censoring when events are not double censored.

sbfnk:
in the bias section.

Closes #82


seabbs (approved):
Nice

PR #226, opened 2024-06-22
Standardise to use natural log

seabbs:
We have a single (?) instance of plotting with log10 and otherwise use the natural log. I think the natural log makes more sense given that is the scale we are modelling on (and teaching epi processes on) but happy with either as long as we stick to one.

### DIFF
--- a/sessions/forecasting-concepts.qmd
+++ b/sessions/forecasting-concepts.qmd
@@ -328,7 +328,7 @@ rw_forecasts |>
   ggplot(aes(x = day)) +
   geom_line(aes(y = .value, group = interaction(.draw, target_day), col = target_day), alpha = 0.1) +
   geom_point(data = example_onset_df, aes(x = day, y = onsets), color = "black") +
-  scale_y_continuous(trans = "log") +
+  scale_y_log() +
   scale_color_binned(type = "viridis") +
   labs(title = "Weekly symptom onset forecasts: log scale",
        col = "Forecast start day")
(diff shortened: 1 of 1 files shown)


sessions/forecasting-concepts.qmd
```diff
@@ -301,7 +301,7 @@ rw_forecasts |>
   ggplot(aes(x = day)) +
   geom_line(aes(y = .value, group = interaction(.draw, target_day), col = target_day), alpha = 0.1) +
   geom_point(data = example_onset_df, aes(x = day, y = onsets), color = "black") +
-  scale_y_log10() +
+  scale_y_continuous(trans = "log") +
```
sbfnk, line 304:
```suggestion
  scale_y_log() +
```
otherwise the y axis will look rubbish

sbfnk (changes requested):

sbfnk (approved):

sbfnk:
Turns out `scale_y_log()` doesn't exist - thinking about this though we're only talking about axis labels, right? I think having 10, 100, 1000 etc. works better than what looks like random numbers.

seabbs:
Then we need to make them all the same and explain why the lines aren't linear on the log10 scale?

sbfnk:
why wouldn't they be, given that changing the base only amounts to a constant scaling on the log scale

seabbs:
> why wouldn't they be, given that changing the base only amounts to a constant scaling on the log scale

Yes your right/

sbfnk:
So can we close this? (and I'd argue do the reverse, i.e. change the plot in forecasting-conectps to log10)

PR #227, opened 2024-06-22
Improve stan output

### LINKED ISSUE #119: update stan blurb
We have a blurb in a few places saying that we add some arguments to suppress output but they should leave them in. Instead we could just suppress output (or do something else).

sbfnk:
Improves the handling of stan sampling statements which now have suppressed outputs rather than random options.

Also adds an explainer of `parallel_chains` and `adapt_delta`. As we're only using `max_treedepth` once I have removed it there rather than explaining.

Closes #119


sbfnk:
> As we're only using max_treedepth once I have removed it there rather than explaining.

Now noticing clash with #212 so will revert.

seabbs (approved):
Nice. Looks good.

PR #231, opened 2024-06-23
Use onset df from a snippet everywhere

### LINKED ISSUE #191: Use onset_df from a snippet everywhere
We currently have a package `example_onset_df` and also use `onset_df` as generated by a snippet of code. We should use the same thing everywhere. The only thing we need to watch for is to make sure that the snippet uses a deterministic process to avoid lots of downstream issues.

kathsherratt:
closes #191

### DIFF
--- a/sessions/forecasting-concepts.qmd
+++ b/sessions/forecasting-concepts.qmd
@@ -294,10 +294,10 @@ Some important things to note about these forecasts:
 - Each forecast used all the data up to the forecast date.
 - We generated 1000 posterior samples for each forecast.
 - We started forecasting 3 weeks into the outbreak, and then forecast once a week (every 7 days), i.e., we created forecasts on day 22, day 29, ... to day 71. We excluded the last 14 days to allow a full forecast.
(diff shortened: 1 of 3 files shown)


sessions/forecasting-concepts.qmd
```diff
@@ -294,10 +294,10 @@ Some important things to note about these forecasts:
 - Each forecast used all the data up to the forecast date.
 - We generated 1000 posterior samples for each forecast.
 - We started forecasting 3 weeks into the outbreak, and then forecast once a week (every 7 days), i.e., we created forecasts on day 22, day 29, ... to day 71. We excluded the last 14 days to allow a full forecast.
-- We used pregenerated simulated data, which is the same as the outbreak we've been using through the course:
+- We used the same simulated outbreak data we've been using through the course:
```
sbfnk, line 297:
```suggestion
- We used the outbreak data simulated in the same way as the one we've been using through the course:
```

sbfnk (changes requested):

sbfnk (approved):

PR #237, opened 2024-06-23
Truncate Gamma priors.

### LINKED ISSUE #197: Are the priors in the `gamma_model.stan` truncated?
In https://github.com/nfidd/nfidd/blob/459d1bfcc5424a1e36f193ce7a575fc8d7ee48cd/sessions/R-Stan-and-statistical-concepts.qmd#L154, you state that the priors are truncated. I thought they would be if they were defined using `T[]` like here https://github.com/nfidd/nfidd/blob/459d1bfcc5424a1e36f193ce7a575fc8d7ee48cd/stan/lognormal.stan#L14.

seabbs:
Closes #197 

I think it isn't totally needed as the lower = 0 already covers you but better practice and I think makes the prior a better representation of what you think

### DIFF
--- a/stan/gamma.stan
+++ b/stan/gamma.stan
@@ -10,7 +10,7 @@ parameters {
 }
 
 model {
-  alpha ~ normal(0, 10);
-  beta ~ normal(0, 10);
+  alpha ~ normal(0, 10) T[0,];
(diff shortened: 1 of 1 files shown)


sbfnk (approved):
I agree

PR #242, opened 2024-06-23
Add missing session introductions

### LINKED ISSUE #144: add 2 sentence plain English summary in the beginning of each session

kathsherratt:
Should close #144


sessions/forecasting-concepts.qmd
```diff
@@ -48,15 +54,6 @@ options(cmdstanr_print_line_numbers = TRUE)
 
 :::
 
-# Introduction to forecasting as an epidemiological problem, and its relationship with nowcasting and R estimation
-
-See the lecture slides.
-
-::: {.callout-tip}
-Proper scoring rules are scoring rules that make sure no model can get better scores than the *true* model, i.e. the model used to generate the data.
```
seabbs, line 56: We need to keep the proper scoring rules definition o think. Ideally we have we are going to use proper scoring rules in the main text and then people can read this to find out what they are

seabbs (approved):
Looks good to me. One comment about where we agreed to put the proper scoring rules definition but I don't feel strongly and as here works.

sessions/forecasting-concepts.qmd
kathsherratt: I agree we keep it, I just shifted it below (line 373) when we start to introduce evaluation. It seemed odd to have it as the first thing in the session

PR #243, opened 2024-06-23
Forecasting models session - fix cutoff to day 71

kathsherratt:
Unless there was a reason to go back to using day 41 in this session (in which case please close this PR)?


seabbs:
We changed it to reduce run time. I'm still looking into what is happening with the AR model as it's kit really working as I thought it would (see the fix ar pr)

PR #244, opened 2024-06-23
end of course slides

### LINKED ISSUE #74: Slides: Summary of the course

sbfnk:
(and some minor updates to intro where stuff was copied from)

Closes #74


seabbs (approved):
This looks good. Only note is that we probably want the session format to be the same as the others i.e the format @kathsherratt has used.

sbfnk:
not sure this is needed here as there is no practical session? Intro doesn't have the same structure either.

PR #247, opened 2024-06-23
More further reading

### LINKED ISSUE #245: Add charniga et al to further reading
https://arxiv.org/abs/2405.08841

### LINKED ISSUE #246: More further reading
On a coach so can't pr myself. 

Jason's paper is a good one for the little bit of mechanism goes a long way

https://pubmed.ncbi.nlm.nih.gov/28342787/

I think we can have one straight up self citation... 

https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1006785

Stoner on nowcasting

https://arxiv.org/pdf/1912.05965

sbfnk:
At some point we might want to add why we chose these but for now I think this is good.

Closes #246 
Closes #245


seabbs (approved):
Yes agree

PR #253, opened 2024-06-23
Echo `source()` code interactively to R notebook

kathsherratt:
Currently we use `source()` so that people can see code, but interactively this doesn't echo the code within the R notebook. This PR adds `echo` and `max_deparse` options to `source()` so that all the code prints directly beneath the call


seabbs:
I thought we would need to use echo everywhere to close this but perhaps I misunderstood?

kathsherratt:
I haven't added echos where we have already introduced these code snippets previously and are just using / manipulating the resulting data - but happy to do so

kathsherratt:
have done so!

seabbs:
ah right got you. I don't have strong feelings either way so which do you prefer?

kathsherratt:
I prefer not to echo code where it's not the focus so have reverted that commit. 
Ready to merge?

seabbs (approved):

PR #255, opened 2024-06-23
Issue 234: Remove all bounds on init_R and remove upper boound on rw_sd

### LINKED ISSUE #234: remove bounds on `init_R` and `rw_sd`
Vague memory is that this is in another issue somewhere but I couldn't find it - if we set the `init` we shouldn't need this (and experimentation in the Rt session confirms we don't).

seabbs:
As the title closes #234


sbfnk:
I see no issue in the render so good to go ahead I think

sbfnk (approved):

PR #256, opened 2024-06-23
Downsample plots

### LINKED ISSUE #217: Number of posterior samples for plotting
Sometimes we plot the whole posterior (eg renewal session), sometimes we sample 100 trajectories (eg convolutions session), sometimes 50 (forecast models session). It's confusing and think we should stay consistent.

kathsherratt:
Should close #217

I notice in the forecasting concepts session some plots are set to [`.draws <= 50`](https://github.com/nfidd/nfidd/blob/61165c416d2b1d12937ea6bf7a34f9a5c180fef6/sessions/forecasting-concepts.qmd#L250). I haven't changed in case there's some reason for selecting the first 50 draws (rather than the random sample used elsewhere)? If no reason I'll update to match.

### DIFF
--- a/sessions/R-estimation-and-the-renewal-equation.qmd
+++ b/sessions/R-estimation-and-the-renewal-equation.qmd
@@ -195,7 +195,7 @@ r_posterior <- r_fit |>
   gather_draws(R[infection_day]) |>
   ungroup() |>
   mutate(infection_day = infection_day - 1) |> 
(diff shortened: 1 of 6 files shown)


seabbs:
>  If no reason I'll update to match.

nope no reason so I think this could do with being updated

seabbs (approved):

PR #260, opened 2024-06-23
Issue 134: Add available data plotting to simple nowcasting session

### LINKED ISSUE #134: Nowcasting session: plot estimates against partial and final data
At the moment we're plotting nowcasts against the final data but don't say that very explicitly.

I think it would be good to also plot against the partial data used for fitting. This could then better motivate departure from the independent onsets model (it does *something* but not very well).

seabbs:
Closes #134 

Adds observed data plotting and a note explaining what is happening. Note I felt I needed to crank up adapt delta for the joint nowcast as it had an unacceptable number of divergent transitions (over 30%). Added a note on this.  I see a runtime of 2 minutes.

### DIFF
--- a/sessions/joint-nowcasting.qmd
+++ b/sessions/joint-nowcasting.qmd
@@ -130,9 +130,7 @@ We also need to update our simulated truth data to include the Poisson observati
 
 ```{r update-simulated-onset}
 noisy_onsets_df <- reporting_triangle |>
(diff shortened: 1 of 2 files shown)


kathsherratt:
This is great!

seabbs:
Okay to merge?

kathsherratt (approved):

Issue #259, opened 2024-06-23
Forecasting plots don't reflect model fit


zsusswein:
@seabbs -- throwing this together quickly to link in a PR.

The differenced AR1 model is producing some weird forecasts in the existing qmd:

![image](https://github.com/nfidd/nfidd/assets/46581799/104fe64a-f4c8-4c15-9ab9-728050d0f1b3)

We did some investigating into source. It seemed like it might be a bug in the Stan code, but after some digging we realized that manually created plots out of the fitted model object look fine:

![image](https://github.com/nfidd/nfidd/assets/46581799/913d0e2b-bd58-4aa2-bf13-0a8311dcf5aa)

![image](https://github.com/nfidd/nfidd/assets/46581799/e5f5e886-2948-4ec2-88c6-83d6c060c667)

It looks to me like the issue is in the plotting code in https://github.com/nfidd/nfidd/blob/fa76691ca77eac23f953a6b043624cf0f657d088/sessions/forecasting-models.qmd#L292-L317

I think there's problems (1) the data being compared isn't lining up properly (join issue) (2) the forecast horizon is so long that the AR process is blowing up will make it hard to see.

I'd propose fixing both by changing the draws extraction and join and limiting the plotting to 55 or so days.

seabbs:
Aweoms thanks @zsusswein. 

Did you find any changes to the priors that made a difference with the model performance?

seabbs:
@kathsherratt do you think we have this problem anywhere else? Like in the forecast concepts session?

seabbs:
ah damn. I think its because the output of `gather_draw` is grouped by `day` and that makes the following filter be grouped by day. That then causes a bunch of partial trajectories?

seabbs:
not clear why this would just be happening here vs everywhere?

zsusswein:
> Did you find any changes to the priors that made a difference with the model performance?

I did play around with the model when I was bug-hunting. I was curious if the constraint on the damping would have caused an issue and I thought the total variance was perhaps too high.

![image](https://github.com/nfidd/nfidd/assets/46581799/8a0fa749-f90a-43eb-a833-571d63498edb)

In the posterior, I don't know if I see that having a large impact. The fitted damping value isn't strongly changed nor is the $R_{t = 0}$. The variance change does have an impact, but it's sort of a dealer's choice situation on what you believe is reasonable. I might make it lower in my own work, but I wasn't planning on modifying it here as a bugfix.

[...]

PR #263, opened 2024-06-23
Add Rt plot and fix forecasting plot

seabbs:
I think this fixes the plotting. Also adds a Rt plot. Changes the prior on the AR model as I think this is more reasonable.


seabbs:
@zsusswein do you think this fixes the plotting issue? 
@kathsherratt I need to go and get on the plane. Could you check our plotting as I think this is a bit of a killer issue! 

If anyone gets the chance I think the AR model should really be doing better than it is score wise. I think this is due to priors? If anyone can investigate and set new ones + update all the fits in data-raw and see if that makes a difference in the session I think that would be great

seabbs:
If not I will try and do that on Monday but it will be difficult communicating why things have changed with participants

zsusswein (approved):
>  do you think this fixes the plotting issue?

LGTM 

image

zsusswein:
You might also want to link #259

seabbs:
Okay to merge @kathsherratt?

kathsherratt (approved):

PR #267, opened 2024-06-23
Every `gather_draws()` followed by `ungroup()`

kathsherratt:
Inspired by #259 and  #263 - consistently use `ungroup()` after every call to `gather_draws()`


kathsherratt:
(FYI the nowcasting sessions and elsewhere used this convention already so everywhere else is OK)

seabbs (approved):
Amazing. I literally cannot believe this is the default!

PR #268, opened 2024-06-23
Reduce prior AR variance

zsusswein:
Per @seabbs request, I'm proposing a tweak to the AR process's prior variance. I think the current process has slightly too much variance and it's hurting the model's predictive performance (Note: I can't get the scores to work in my installation due to some kind of a bug in loading scoringutils). 

After the change, the prior predictive Rt looks like so: 

image

The posterior for Rt looks like:

image

In the current version on main the posterior looks like:

image

To my eye, the current posterior looks like it's shunting real changes in epidemic dynamics from the mean trend into the variance. I think that problem is rectified by the reduction in the prior AR variance, but fundamentally this prior is a subjective judgement call.

### DIFF
--- a/stan/statistical-r.stan
+++ b/stan/statistical-r.stan
@@ -36,7 +36,7 @@ model {
   // priors
   init_R ~ normal(-.1, 0.5); // Approximately Normal(1, 0.5)
   rw_noise ~ std_normal();
-  rw_sd ~ normal(0, 0.05) T[0,];
+  rw_sd ~ normal(0, 0.01) T[0,];
   damp ~ normal(0.5, 0.25) T[0, 1];
   obs ~ poisson(onsets[1:n]);
 }
--- a/stan/statistical-r.stan
+++ b/stan/statistical-r.stan
@@ -36,7 +36,7 @@ model {
   // priors
   init_R ~ normal(-.1, 0.5); // Approximately Normal(1, 0.5)
   rw_noise ~ std_normal();
-  rw_sd ~ normal(0, 0.05) T[0,];
+  rw_sd ~ normal(0, 0.01) T[0,];
   damp ~ normal(0.25, 0.25) T[0, 1];
   obs ~ poisson(onsets[1:n]);
 }


zsusswein:
FYI @seabbs @kathsherratt @sbfnk -- opened as requested. I don't have the ability to assign or request reviewers, so I leave this in your capable hands from here.

kathsherratt:
Thanks so much @zsusswein ! What did you think about the suggested [change to damp](https://github.com/nfidd/nfidd/pull/263/files#diff-c8be9d8ceb5d61c49584de7aafa6abbb75ea20be330596b248ecd5e0337aa006R40) in #263 in this case?

Flagging for review by @sbfnk as well if possible.

zsusswein:
I don't have great intuition about it, but I hacked together a super quick little simulation in R to take a look:

#### damp ~ N(0.5, 0.25)

image

#### damp ~ N(0.25, 0.25)

image

Looks to me like the impact of switching the prior damping mean to 0.5 is pretty minimal, but totally reasonable. The simulation is rough though, so I wouldn't take it too seriously.

kathsherratt:
Thank you so much again, this is super helpful!

Now merged #263, but my apologies for the merge conflict here (I'm happy to resolve if you're ok for me to push to `main` on your fork).

The change you suggest here produces better results for me too. I think let's go ahead with this and I'll update the saved forecasts.

zsusswein:
Great! Good luck with the course.

kathsherratt (approved):

seabbs:
The thing we now need is the saved forecasts in data raw for the AR model to updated. Anyone on that?

kathsherratt:
yes I am on that now.

seabbs:
The long to fit models might take a while. If that is the case it might need to be run on a big box somewhere. Mine is having some kind of internet issues so cant do remotely 

Ideally this will mean improved performance. If it does that might mean we need to make edits to the solutions in the session. Happy to do that in the morning.

seabbs:
Am I right that we already have everything here on main and so should close with many thanks for the heads up? I

PR #271, opened 2024-06-23
Renewal remove forecast

### LINKED ISSUE #127: Forecasting in the renewal eqn session

sbfnk:
splits the model in two

Closes #127


seabbs (approved):
This looks okay. Not sure what the direct zero modek is for

sbfnk:
me neither - will remove

PR #276, opened 2024-06-23
Geometric rw functions

### LINKED ISSUE #122: ensure we have an R function for every stan function

sbfnk:
adds R functions for the geometric random walks

@seabbs if you don't like this in the forecasting session I can take it out again

Closes #122

### DIFF
--- a/sessions/forecasting-models.qmd
+++ b/sessions/forecasting-models.qmd
@@ -199,16 +199,16 @@ $$
 ::: {.callout-note, collapse = TRUE}
 ## What behaviour would we expect from this model?
 
+Again we look at an R function that performs this random walk
+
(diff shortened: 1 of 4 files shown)


sessions/forecasting-models.qmd
```diff
@@ -199,18 +199,19 @@ $$
 ::: {.callout-note, collapse = TRUE}
 ## What behaviour would we expect from this model?
 
+Again we look at an R function that performs this random walk
```
seabbs, line 202: Autoregressive not ransom walk. Can be fixed in another pr

seabbs (approved):

sessions/forecasting-models.qmd
sbfnk: have rephrased

PR #287, opened 2024-06-24
update forecast models

kathsherratt:
Updated forecasts after #207 

Unfortunately this looked like it was improving scores a little bit for the single model but barely changes across the whole. Overall CRPS (natural scale) 65 to 63; or on the log scale, 0.55 to 0.53.

#### Forecasts

Before

image

After

image

#### PIT (log transform)

Before

image

After

image


seabbs (approved):
Hmm how very odd. Still at least we don't need to update the solutions.. 

Thanks for doing that!

PR #354, opened 2024-10-25
Move to a package structure for R and stan code

seabbs:
- Moves R functions into an R folder, adds roxygen
- Ports snippets into R functions, adds roxygen
- Ports stan code to inst
- Adds R side tools for handling stan functions and models
- Uses all of above in course material

It also: 

- Updates to the latest version of scoringutils
- Added a scheduled build for the course so we know it is working over time
- Fixes some minor linting and line length issues

It does not significantly adapt sessions to account for the new package structure. It might be the case there are rough edges that we need to smooth out.


seabbs:
I've successfully built this locally

seabbs:
The failure in the CI makes me think there is an issue with the cmdstan install. I've cleared the cache and rerun but that doesn't seem to be helping.

seabbs:
Confirmed the issues aren't localised to the delay distributions example by `eval = FALSE`

seabbs:
🚀

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: nfidd
 Title: Material to support course on nowcasting and forecasting of infectious
     disease dynamics
-Version: 0.0.0.9999
+Version: 0.1.0.9999
```
sbfnk, line 4: shall we release the current version on `main` as 0.1.0 then before merging this?

R/nfidd-stan-tools.r
```diff
@@ -0,0 +1,304 @@
...
+#'
+#' @return A CmdStanModel object.
+#'
+#' @export
+#'
+#' @family modelhelpers
+#'
+#' @examplesIf requireNamespace("cmdstanr", quietly = TRUE)
+#' if (!is.null(cmdstanr::cmdstan_version(error_on_NA = FALSE))) {
+#'   model <- nfidd_cmdstan_model("renewal", compile = FALSE)
```
sbfnk, line 279:
```suggestion
#'   model <- nfidd_cmdstan_model("simple-nowcast", compile = FALSE)
```
or any other one that actually exists?

R/nfidd-stan-tools.r
```diff
@@ -0,0 +1,304 @@
...
+    include_paths = nfidd::nfidd_stan_path(),
+    ...) {
+  if (!requireNamespace("cmdstanr", quietly = TRUE)) {
+    stop("Package 'cmdstanr' is required but not installed for this function.")
+  }
+
+  stan_model <- system.file(
+    "stan", paste0(model_name, ".stan"),
+    package = "nfidd"
+  )
```
sbfnk, line 293: This doesn't work with `devtools::load_all()`. Perhaps not the end of the world but could be fixed by instead using 
```suggestion
  stan_model <- fs::path_package(
    "nfidd", "stan", paste0(model_name, ".stan")
  )
```

sessions/R-Stan-and-statistical-concepts.qmd
```diff
@@ -28,20 +28,19 @@ The source file of this session is located at `sessions/R-Stan-and-statistical-c
...
 
-In this session we will use the `ggplot2` library for plotting, the `here` library to reference files, for example, the stan model, and the `cmdstanr` library for using stan.
+In this session we will use the `ggplot2` library for plotting, the `nfidd` package to load the stan models, and the `cmdstanr` library for using stan.
 We will also use the `bayesplot` and `posterior` packages to investigate the results of the inference conducted with stan.
 
 ```{r libraries, message = FALSE}
+library("nfidd")
 library("ggplot2")
-library("here")
 library("cmdstanr")
```
sbfnk, line 37: do we still need these or is this handled by `nfidd` now?

DESCRIPTION
```diff
@@ -18,20 +18,23 @@ BugReports: https://github.com/nfidd/nfidd/issues
...
 Remotes:
-    stan-dev/cmdstanr,
-    epiforecasts/scoringutils@d3cf08d
+    epiforecasts/scoringutils@b370557
+Additional_repositories:
+    https://stan-dev.r-universe.dev
 Encoding: UTF-8
 Language: en-GB
+RoxygenNote: 7.3.2
+LazyData: true
```
sbfnk, line 40:
```suggestion
LazyData: true
KeepSource: true
```
so that people can see comments inside the functions

sbfnk (changes requested):
Great! I think this will work much much better than what we had before.

I didn't review all the functions butI skimmed the rendered sites and all looks good with results showing as expected. As you say a follow-up PR should tidy up the text and clarify the role of the package etc. to make sure it all make sense and is consistent.

R/nfidd-stan-tools.r
seabbs: Oh weird it works for me and I use it as a pattern everywhere?

sessions/R-Stan-and-statistical-concepts.qmd
seabbs: We could move it to an imports?

R/nfidd-stan-tools.r
seabbs: If changing we need to add to the description

sbfnk: > If changing we need to add to the description

Yes. I'm OK to leave as is for now and handle via `devtools::install()` and reloading if/where needed.

sessions/R-Stan-and-statistical-concepts.qmd
sbfnk: My question was more whether we're actually using any `cmdstanr` functions anywhere. No harm in calling the library I guess, but was wondering whether we strictly still need to.

sbfnk (approved):

PR #355, opened 2024-10-26
Ensemble session

seabbs:
Adds an ensembling session based on our current forecasting models session. WIP.


seabbs:
ive got the barebones in here but it needs more context especially in the qra section. i think it communicates the challenges flagged in the slides

sbfnk:
This looks really good. I'd love to see how the stackr mixture ensemble performs but that might be too much for this session. I'll leave some edits in a separate PR.

seabbs:
Yeah I agree I thought about adding but didn't get a chance. We could put it in an extra work section if we think it's too much for the main session?

PR #357, opened 2024-10-31
add mixture ensembles

sbfnk:
I think it makes a nice addition for the "full" version but it makes an already quite complicated session even more complicated. Perhaps something we want to cut again when thinking about this for a 60 minute session.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,8 @@ Suggests:
     rmarkdown,
     scoringutils,
     tidybayes,
-    qra
+    qra,
+    stackr
 Remotes:
     epiforecasts/scoringutils@b370557,
     epiforecasts/qra@e86d7e0,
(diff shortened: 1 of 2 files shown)


seabbs (approved):
From a site read this looks great. I agree for Idm we will need to think about timing. I see an install error on deployment that looks to be coming from Stacie

sbfnk:
Yes, and I don't really understand it. Am investigating on the side as not a priority for now.

seabbs:
I had a look and I don't either unless it's something super obvious like a misnamed model being imported by restan which seems unlikely

sbfnk:
Turned out this fixed it. Not sure how this ended up in Rbuildignore. 

https://github.com/epiforecasts/stackr/commit/1853faf5a27df948706b229bb358f9ea268c7d24

seabbs:
I'm on my phone so can't resolve merge issues but otherwise this looks good

PR #429, opened 2025-06-19
sort out check issues

sbfnk:
closes #388 
closes #422 
closes #423 

and fixes a few remaining check issues (license files)
and fixes a formatting problem with in the documentation of the summarise_lognormal() function

I suspect we may still have to wait until lopensemble is in the universe until we pass checks.

PR #430, opened 2025-06-19
make lintr happy

sbfnk:
ignored summarised_lognormal as that's updated in another PR


sbfnk:
linting succeeds but will have to wait for #429 for checks to pass

PR #460, opened 2025-06-20
split session 8

### LINKED ISSUE #458: Split forecasting concepts into two
i.e into visualisation and evaluation.

sbfnk:
closes #458

### DIFF
--- a/reference/sessions.qmd
+++ b/reference/sessions.qmd
@@ -86,17 +86,23 @@ Tuesday June 24: 11.00-12.30
     - joint estimation of delays, nowcasts and reproduction numbers (35 mins)
 -   Wrap up (10 mins)
 
-## Session 8: Forecasting concepts
+## Session 8: Visualising infectious disease forecasts
 
-Tuesday June 24: 14.00-15.30
+Tuesday June 24: 14.00-14.45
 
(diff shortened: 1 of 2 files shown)


reference/sessions.qmd
```diff
@@ -86,17 +86,23 @@ Tuesday June 24: 11.00-12.30
     - joint estimation of delays, nowcasts and reproduction numbers (35 mins)
 -   Wrap up (10 mins)
 
-## Session 8: Forecasting concepts
+## Session 8: Visualising infectious disease forecasts
```
seabbs, line 89: do we actually want to keep this called forecast concpets and rename the session? It feels more like forecasting concepts than vis imo?

seabbs (approved):
I think we need to change the yaml order and would think about changing the name of the forecast vis session to be a bit more in line with key content.

We also need to update the learning objectives

reference/sessions.qmd
sbfnk: yep sounds good

sbfnk:
> I think we need to change the yaml order and would think about changing the name of the forecast vis session to be a bit more in line with key content.

will make a separate Issue as I think there's a slightly broader point

sbfnk:
> We also need to update the learning objectives

They're not by session anyway - but I've added one on visualisation

seabbs (approved):

PR #463, opened 2025-06-20
call variable `reported_onsets`

### LINKED ISSUE #307: consistent use of `obs` or `reported_onsets`
NBD but when we add poisson uncertainty in the early sessions we name this variable `observed` to differentiate it from `onsets`. In the nowcasting sessions we use `reported_onsets` to do the same thing. I think we should pick one and use it consistently. (I'd probably pick using `reported_onsets` in the earlier ones.)

sbfnk:
closes #307


seabbs (approved):
LGTM

PR #468, opened 2025-06-20
try to better explain convolutions

### LINKED ISSUE #295: clarify lecture slides on convolutions
the discrete -> continuous steps turned out quite hard to explain from the slides - equation on first slide is new and, I think, unneeded.

sbfnk:
Closes #295 

Or at least attempts to do so - not sure if this is clearer but perhaps it is


sbfnk:
or actually just the From individual events to population level counts slide

seabbs:
> or actually just ditch the From individual events to population level counts slide

Yes I mean I am clearly not following. I think this option

Issue #409, opened 2025-06-17
Session Review: Delay Distributions


seabbs:
Review the delay distributions session content and slides.

## Key areas to review:
- Mathematical concepts explanation
- Practical examples and intuition
- R code demonstrations
- Slide clarity for complex distributions
- Exercise difficulty and progression

sessions/delay-distributions.qmd

Issue #390, opened 2025-06-06
review "going further" sections


kathsherratt:
- Add examples from real-world applications
- Add any more/less established "best practices"

seabbs:
Shall we move this into something to do in each section review?

sbfnk:
sounds good to me

seabbs:
Closing then with this in mind

Issue #387, opened 2025-06-06
add references


kathsherratt:
include references in each session (adopting/expanding bib file from the IDM material)

PR #469, opened 2025-06-20
session 2 going further

kathsherratt:
#390 
#409 

- [x] ref to be updated later with general bib file update #387


sessions/delay-distributions.qmd
```diff
@@ -236,8 +236,23 @@ ggplot(res_df, aes(x = x, y = density, group = .draw)) +
...
+## Challenge
+
+-   In this session we were in the enviable situation of knowing which distribution was used to generate the data. 
+With real data, of course, we don't have this information available. 
+Try using a different distribution for inference (e.g. normal, or gamma). 
+Do you get a good fit?
+
+## Methods in practice
+
+-   Charniga et al., [Best practices for estimating and reporting epidemiological delay distributions of infectious diseases using public health surveillance and healthcare data](https://doi.org/10.48550/arXiv.2405.08841) summarises challenges in estimating delay distributions into a set of best practices, with a practical flowchart and reporting checklist.
```
seabbs, line 248: do we want this here or in the biases one?

sessions/delay-distributions.qmd
```diff
@@ -236,8 +236,23 @@ ggplot(res_df, aes(x = x, y = density, group = .draw)) +
...
 
 # Wrap up
 
+The learning objectives for this session were:
```
seabbs, line 252: Same point

sessions/delay-distributions.qmd
```diff
@@ -236,8 +236,23 @@ ggplot(res_df, aes(x = x, y = density, group = .draw)) +
 
 # Going further
 
--   In this session we were in the enviable situation of knowing which distribution was used to generate the data. With real data, of course, we don't have this information available. Try using a different distribution for inference (e.g. normal, or gamma). Do you get a good fit?
+## Challenge
+
+-   In this session we were in the enviable situation of knowing which distribution was used to generate the data.
```
seabbs, line 241: How does this render in markdown looks a bit odd?

sessions/delay-distributions.qmd
kathsherratt: I was adding here to flag it early on, but don't feel very strongly

seabbs: I think we should push to bias if that is okay as otherwise the content will all be new to people

seabbs (approved):

Issue #442, opened 2025-06-19
Session Review: Using Delay Distributions to Model Data Generating Process


seabbs:
Review and improve Session 4: Using delay distributions to model the data generating process of an epidemic covering:
- Introduction to using delay distributions to model data generating process
- Implementing convolution models
- Identifying potential problems with convolution approaches

This is part of the systematic session review process.

seabbs:
Related specific issues to address during this review:
- #295: clarify lecture slides on convolutions
- #298: explain convolution with a worked example
- #299: discrete counts shown as continuous lines on convolution slides
- #300: convolve_with_delay indexing issue

PR #471, opened 2025-06-20
session-4-going-further

kathsherratt:
#390 / #442


sessions/using-delay-distributions-to-model-the-data-generating-process-of-an-epidemic.qmd
```diff
@@ -315,8 +317,20 @@ This is particularly useful when dealing with arrays such as `inference` because
     How would this look in stan code?
     Would you expect it to yield a different result?
 
+## Methods in practice
+
```
seabbs, line 321: EpiNow2 data simulator, simulist?

Issue #410, opened 2025-06-17
Session Review: R Estimation and Renewal Equation


seabbs:
Review the R estimation session content and slides.

## Key areas to review:
- Renewal equation explanation and intuition
- Mathematical derivations vs practical understanding
- Stan model implementation walkthrough
- Slide balance between theory and practice
- Exercise complexity for target audience

sessions/R-estimation-and-the-renewal-equation.qmd

seabbs:
Related specific issues to address during this review:
- #318: turn generated quantities into R

PR #472, opened 2025-06-20
session-5-going-further

kathsherratt:
#390 / #410


seabbs (approved):
LGTM. A concern about duplicating the learning points

kathsherratt:
shared concern but if so let's drop the "Wrap up" header as this is empty?

seabbs:
we had this here as a marker for us and I think that works. I suggest linking to the learning objectives and maybe we have some boiler plater about reviewing what you have learnt?

sessions/R-estimation-and-the-renewal-equation.qmd
```diff
@@ -486,11 +487,28 @@ We can see that the estimates are smoother when using the random walk model for
...
+
 - We have used symptom onsets under the assumption that every infected person develops symptoms. 
 Earlier we also created a time series of hospitalisation under the assumption that only a proportion (e.g., 30%) of symptomatic individuals get hospitalised. 
 How would you change the model in this case?
 What are the implications for inference?
-- The [`EpiNow2`](https://epiforecasts.io/EpiNow2/) package implements a more complex version of the model we have used here. You will have the opportunity to try it in a later session.
+
+## Methods in practice
+
+- The `EpiEstim` software provides a range of accessible tools for estimating $R_t$, using a simpler approach than the model we have used here. Alternatively, the `EpiNow2` [@abbottEpiNow2EstimateRealtime2025] package implements a more complex version of the model we have used. Taking a different approach, `Epidemia` @bhattSemiMechanisticBayesianModeling2020 implements a regression model approach to $R_t$ estimation.
```
seabbs, line 499:
```suggestion
- `EpiEstim` provides a range of accessible tools for estimating $R_t$, using a simpler approach than the model we have used here.
- `EpiNow2` [@abbottEpiNow2EstimateRealtime2025] package again implements a range of models similar to the model we have used here.
- `Epidemia` @bhattSemiMechanisticBayesianModeling2020 implements a regression model approach to $R_t$ estimation.
```

sessions/R-estimation-and-the-renewal-equation.qmd
```diff
@@ -486,11 +487,28 @@ We can see that the estimates are smoother when using the random walk model for
...
+- In this course we focused on the instantaneous reproduction number, which is most suitable for real-time analysis when not all infections have yet been observed. Methods to estimate the case reproduction number are implemented in e.g. @wallingaDifferentEpidemicCurves2004, and are most appropriate using retrospective data that represent a complete final count of an outbreak.
```
seabbs, line 500:
```suggestion
- In this course we focused on the instantaneous reproduction number.
An alternative is the case reproduction number implemented in @wallingaDifferentEpidemicCurves2004.
```
This is often said but I think it's really quite a lot more nuanced than this so removing.

seabbs (approved):

sessions/R-estimation-and-the-renewal-equation.qmd
```diff
@@ -1,5 +1,6 @@
 ---
 title: "$R_t$ estimation and the renewal equation"
+bibliography: ../nfidd.bib
```
seabbs, line 3:
```suggestion
```

Removing for now and need to add back in in #491

seabbs (approved):

PR #475, opened 2025-06-20
fix convolution function

### LINKED ISSUE #300: convolve_with_delay seems to have indexing wrong
double check

sbfnk:
Looking again I think that indeed the indexing was going wrong.

Closes #300.

Unfortunately this means once again re-generating the data. Doing it now so don't merge yet please.


seabbs (approved):
Not ideal and agree with the fix. Am I correct in thinking this should only have impacted the early parts of the time series?

sbfnk:
That matches my understanding, yes

Issue #336, opened 2024-06-26
add glossary


sbfnk:

sbfnk:
As a alternative, we could provide a quarto search? This might be preferable as it doesn't require separate maintenance, and agreeing on the right shape for a glossery in #490 has proven tricky.

PR #490, opened 2025-06-22
Glossary

kathsherratt:
#336
possible problems?
- might be too long
- have arranged by concept rather than alphabetical


about/glossary.qmd
```diff
@@ -0,0 +1,65 @@
...
+A focus of this course is on understanding the time between these pairs of events.
+Key pairs of events include: 
+- The incubation period: infection to symptom onset 
+- Symptom onset to hospitalisation 
+- The generation time: infection in one person causing infection in another
+
+### Session 3
+
+**Censoring**: knowing that an event occurred at some time, but not specifically when.
+In this course, we mostly consider double (or interval) censored events: we know the event occurred at some time within a time interval, such as at any time on a certain date.
```
seabbs, line 28:
```suggestion
In this course, we mostly consider double (interval) censored events: we know that both events occurred at some time within a time interval, such as at any time on a certain date.
```

They are double interval censored.

about/glossary.qmd
```diff
@@ -0,0 +1,65 @@
...
+Events might also be right censored (we only know the earliest time it could have happened) or left censored (we only know the latest time it could have occurred).
```
seabbs, line 29: I think these are the wrong way around? I think if something is right censored we know events up until a latest time and then beyond that we know they happened but not the value.

about/glossary.qmd
```diff
@@ -0,0 +1,65 @@
...
+
+**Right truncation**: when we only have incomplete information about events because not enough time has passed to observe all of them.
```
seabbs, line 31: It would be nice to standardise the here as truncation has the same flavours as censoring with the added flavour of we also don't k ow if events occurred. I think the Charniga paper has some defs we can lift has these things are tricky to talk about.

seabbs: By standardise I mean introduce censoring and truncation then types I.e right left interval and the the idea of double which is just where both events in a delay have some censoring

about/glossary.qmd
```diff
@@ -0,0 +1,65 @@
...
+
+**Discretisation**: when we want to use a continuous probability distribution to model a discrete (count) variable, we need to transform the probability distribution from continuous to discrete.
+
+Moving back and forth between continuous and discrete / simulation and estimation
+
+### Session 5
+
+**Reproduction number**: the average number of secondary infections caused by a single infectious individual and can in time and space as a function of differences in population level susceptibility, changes in behaviour, policy, seasonality etc.
+
+**Renewal equation**: relates the number of infected people in the population; the current value of the reproduction number; and a delay distribution of the so-called generation time
```
seabbs, line 46:
```suggestion
**Renewal equation**: relates the number of infected people in the population; the current value of the reproduction number; and a delay distribution of the generation time
```

about/glossary.qmd
```diff
@@ -0,0 +1,65 @@
...
+
+**Random walk**: random noise along a given trajectory. In this course, we use a **geometric random walk**, where we add this random noise the logarithm of the reproduction number.
```
seabbs, line 48:
```suggestion
**Random walk**: random noise along a given trajectory. In this course, we use a **geometric random walk**, where we add this random noise to the logarithm.
```

seabbs (commented):
I did a short read through and then got confused that there were two? I'm not sure how much content we want to put in here? My slight concern is this is tending towards being the whole course!

kathsherratt:
> I did a short read through and then got confused that there were two? 
Sorry, my bad! Fixed

> I'm not sure how much content we want to put in here? My slight concern is this is tending towards being the whole course!
Yeah agreed, it's too much, very happy if you suggest mass deletions, I was erring on too much than too little

seabbs:
I am wondering if we want to stick on this until after the course as it might take a fair amount of work.

sbfnk:
Are there still plans to pursue / maintain this?

sbfnk:
Closing for now as stale - do open again if more work on this is planned.

PR #492, opened 2025-06-22
update ensemble session

sbfnk:
This:
- updates the structure to give equal balance to quantile-based and sample-based ensembles, at least in the structure if not in content
- updates references to `stackr` to `lopensemble`
- fixes a bug in the inverse WIS weighting where WIS values were not considered for cumulatively increasing historical data, but only for the date of the forecast (NB the fact that this didn't perform better might hint that this is a bad idea to begin with)
- updates the solutions/interpretations in line with these fixes and https://github.com/epiforecasts/lopensemble/pull/40, especially the distribution of weights, which looks a lot more like what one would expect:

image

At a later stage I think this session could do with a broader rewrite. It is very long and could probably be split into two. Looking at unweighted/filtered/inverse CRPS weighted mixture ensembles would also make sense for comparison with the quantile version.

### DIFF
--- a/sessions/forecast-ensembles.qmd
+++ b/sessions/forecast-ensembles.qmd
@@ -545,9 +545,10 @@ In order to perform inverse WIS weighting we first need to calculate the WIS for
 model_scores <- quantile_forecasts |>
   score()
 cumulative_scores <- list()
+## filter for scores up to the origin day of the forecast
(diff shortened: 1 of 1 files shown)


seabbs (approved):
Nice looks good!

PR #500, opened 2025-06-23
update install instructions

sbfnk:
to use remotes and r-universe

I think this is the only option that 
- updates dependencies/suggests
- uses universe


seabbs (approved):

seabbs:
What about #500 ? We just need a universe release for the course?

PR #540, opened 2025-06-27
updating page names and ordering of Thursday's Forecasting modules.

nickreich:
I took a first stab at re-organizing the content for Thursday's forecasting sessions. The ordering is now

 - forecast-visualization --> epi-forecasting-concepts
 - forecasting-models
 - forecast-evaluation
 - more-complex-forecasting-models (NEW)
 - forecast-ensembles

With the modules

- examples-of-available-tools
- methods-in-the-real-world

reordered to be at the end.

I also deleted the folders with figures for a few of the renamed folders.


seabbs:
Looks good. I think this needs to point at the sismid repo?

PR #634, opened 2026-06-21
Correct overstated bias claim for ad-hoc censoring fixes

seabbs:
## ⚠️ DRAFT — statistical correction needing maintainer/expert review

This PR rewords a statistical claim in `sessions/biases-in-delay-distributions.qmd`. **A human maintainer with statistical expertise should verify before merging.** The change was prompted by a reviewer suspicion that the original sentence was wrong, and the evidence below (Park et al. 2024 + a fresh simulation) supports that suspicion.

## The claim under question

Previously (around line 245-247):

> There are many ad-hoc solutions to this problem that ... introduce a shift in the data by half a day ... or discretise the distribution and use the difference between two cumulative density functions with a day, or two day interval.
> **Of these all but the two-day interval approach introduce more bias than doing nothing as above.**

The bolded sentence claims that, of the ad-hoc fixes, *only* the two-day interval beats the naive ("do nothing") approach. This is too strong and is not what the evidence supports.

## What Park et al. 2024 actually says

Park et al. (`parkEstimatingEpidemiologicalDelay2024`, medRxiv 10.1101/2024.01.12.24301247) is the cited reference. Its relevant statements:

- The **correct** discretised double-censoring likelihood for daily reporting is a **two-day window**: `L ∝ F(d+1) − F(d−1)` (their Eq. 35). "Rephrasing the likelihood in this way highlights that the correct censoring interval for this approach is `[d−1, d+1]`."
- The **single-day** `[d, d+1]` interval "in effect only accounts for the censoring of a single event and hence **induces additional bias beyond that introduced by assuming a uniform distribution**" — i.e. it is worse than the *correct two-day* approach.
- The **midpoint** `[d−0.5, d+0.5]` "assumes the primary event is known to have taken place at the midpoint of the interval, and that all primary and secondary event times are known."
- The two-day interval-reduced approach **performs well** ("gives only slightly more biased estimates" than the best latent-variable method).
- On the naive method: "the naive method that accounts for no biases **outperformed** [single-bias] methods as censoring and truncation biases partially cancelled each other when growth was fast."

Crucially, **Park et al. never claims the half-day shift or single-day discretisation is more biased than the naive/do-nothing approach.** Its comparisons are against the *correct* two-day approach, not against naive. So the original sentence overstates and mis-attributes the paper's findings.

## Simulation evidence

Self-contained base-R simulation (no Stan). Continuous primary event times (with growth rate `r`), continuous lognormal delay (`meanlog=1.0, sdlog=0.5`, true mean = 3.080 — the session's own parameters), **both** event times floored to integer days (double interval censoring), observed integer delay `d`. Lognormal MLE under each method; `N = 2e5`.

`|bias| vs naive` is `|bias_method| − |bias_naive|`: **negative = less biased than doing nothing, positive = more biased.**

### Session parameters (true mean 3.080)

| method | est. mean | bias | flat r=0 | growing r=0.15 | declining r=−0.15 |
|---|---|---|---|---|---|
| naive (`d + 0.01`) — *do nothing* | 3.23 | +0.15 | 0.000 | 0.000 | 0.000 |
| shift +0.5 (`d + 0.5`) | 3.58 | +0.50 | **+0.355** | **+0.368** | **+0.349** |
| single-day forward `F(d+1)−F(d)` | 3.58 | +0.50 | **+0.347** | **+0.360** | **+0.341** |
| single-day **centred** `F(d+0.5)−F(d−0.5)` | 3.09 | +0.01 | −0.139 | −0.126 | −0.141 |
| two-day `F(d+1)−F(d−1)` | 3.07 | −0.01 | −0.139 | −0.144 | −0.122 |

### Robustness (flat unless noted)

- Longer delay (true mean 6.52): same pattern — `shift+0.5` and `single-day forward` are ~+0.47 worse than naive; `centred` and `two-day` ≈ unbiased. Holds at r=0.3.
- Shorter delay (true mean 1.87, comparable to the 1-day interval — the high-risk regime the session flags): naive bias balloons to +0.62 (33%), and **every** ad-hoc method (including shift+0.5 and single-day forward) is *less* biased than naive.

## Verdict

The original sentence is **wrong / overstated**:

1. The **centred one-day** CDF difference `F(d+0.5)−F(d−0.5)` — a natural reading of "difference between two CDFs with a one-day interval" — is essentially **unbiased**, as good as the two-day window and far better than naive. So it is false that "all but the two-day interval" are worse than naive.
2. Only the **half-day data shift** and the **single-day forward** difference `F(d+1)−F(d)` are reliably worse than naive, and even that flips for short delays.
3. The relative ranking depends on delay length and growth rate, so a blanket statement is not defensible.

## The reword

The new text explains *why* two days is the correct window (one day of uncertainty per censored event → `F(d+1)−F(d−1)`), notes that single-day-forward and half-day-shift can be worse than doing nothing, that the centred one-day and two-day windows are both near-unbiased, and that the ranking is regime-dependent — steering readers to the principled double interval censoring model rather than any ad-hoc fix. It keeps the Park et al. citation.

### Notes for reviewer
- I did not run `quarto render`. Please confirm the inline `$...$` math renders.
- The naive bias is positive and modest at the session's 3-day mean; the simulation uses `floor()` on both events to match the session's data-generating code exactly.
- Open question for the author: the original sentence may have intended a *specific* operationalization (e.g. single-day = forward `[d,d+1]` only). If so, the wording should still be corrected because the current text reads as a blanket claim and cites Park et al. for it, which the paper does not support.

### DIFF
--- a/sessions/biases-in-delay-distributions.qmd
+++ b/sessions/biases-in-delay-distributions.qmd
@@ -242,8 +242,11 @@ res |>
 Usually the estimates will be further from the "true" parameters than before when we worked with the unrounded data.
 :::
 
-There are many ad-hoc solutions to this problem that, for example, introduce a shift in the data by half a day to centre it on mid-day or discretise the distribution and use the difference between two cumulative density functions with a day, or two day interval.
-Of these all but the two-day interval approach introduce more bias than doing nothing as above.
+There are many ad-hoc solutions to this problem that, for example, introduce a shift in the data by half a day to centre it on mid-day or discretise the distribution and use the difference between two cumulative density functions with a one-day, or two-day interval.
+The key point is that the observed integer delay corresponds to *two* days of combined uncertainty (one day from each censored event), so the correct discretised likelihood compares the cumulative distribution function across a two-day window, $F(d+1) - F(d-1)$ [@parkEstimatingEpidemiologicalDelay2024].
+Approximations that only span a single day, such as the forward difference $F(d+1) - F(d)$ that in effect accounts for the censoring of just one event, can therefore be *more* biased than doing nothing as above, and a half-day shift of the data tends to over-correct in a similar way.
+By contrast, a centred one-day difference $F(d+0.5) - F(d-0.5)$ and the two-day window are both close to unbiased.
+The relative ranking of these approximations depends on the length of the delay and the epidemic growth rate, so rather than relying on any of these ad-hoc fixes we recommend the principled double interval censoring approach described below.
 See the [How epidemic dynamics affect bias severity](#how-epidemic-dynamics-affect-bias-severity) callout for a visual comparison of these approaches and Park et al. for a detailed discussion of approximate approaches [@parkEstimatingEpidemiologicalDelay2024].
 
 To properly account for double interval censoring, we need to modify the model to include the fact that we don't know when exactly on any given day the event happened.
--- a/sessions/biases-in-delay-distributions.qmd
+++ b/sessions/biases-in-delay-distributions.qmd
@@ -243,10 +243,7 @@ Usually the estimates will be further from the "true" parameters than before whe
 :::
 
 There are many ad-hoc solutions to this problem that, for example, introduce a shift in the data by half a day to centre it on mid-day or discretise the distribution and use the difference between two cumulative density functions with a one-day, or two-day interval.
-The key point is that the observed integer delay corresponds to *two* days of combined uncertainty (one day from each censored event), so the correct discretised likelihood compares the cumulative distribution function across a two-day window, $F(d+1) - F(d-1)$ [@parkEstimatingEpidemiologicalDelay2024].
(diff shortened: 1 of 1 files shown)


sessions/biases-in-delay-distributions.qmd
```diff
@@ -242,8 +242,11 @@ res |>
 Usually the estimates will be further from the "true" parameters than before when we worked with the unrounded data.
 :::
 
-There are many ad-hoc solutions to this problem that, for example, introduce a shift in the data by half a day to centre it on mid-day or discretise the distribution and use the difference between two cumulative density functions with a day, or two day interval.
-Of these all but the two-day interval approach introduce more bias than doing nothing as above.
+There are many ad-hoc solutions to this problem that, for example, introduce a shift in the data by half a day to centre it on mid-day or discretise the distribution and use the difference between two cumulative density functions with a one-day, or two-day interval.
```
seabbs, line 245: this is way to much info. We just need to reword the exsiting sentence vs info dump

seabbs (commented):
LGTM

sbfnk (approved):

PR #639, opened 2026-06-21
Plot the forecast reproduction number over the forecast horizon

seabbs:
In the forecasting concepts session, the plot under "We can now plot the forecast and estimated reproduction numbers" did not show the forecast Rt extending past the cutoff.

**Cause:** the forecast model (`estimate-inf-and-r-rw-forecast.stan`) computes the reproduction number over the forecast horizon (`f_R`) in `generated quantities`, but only returned the forecast *onsets*. So `gather_draws(R[day])` on the forecast fit only covered days up to the cutoff, and the blue forecast Rt line stopped at the vertical dashed line.

**Fix:**
- Output `forecast_R` (the reproduction number over the forecast horizon) from the model.
- Append it to the forecast Rt trajectory in the `extract-Rt` chunk (shifted by `cutoff`) so the forecast extends across the horizon, matching the onset forecast plots.

Verified locally: the model compiles, and an end-to-end fit confirms `forecast_R` now spans the full 28-day horizon (days 72–99) with finite values. `_freeze/` is gitignored so CI renders fresh. The model is only used in this session.

> Generated by an agent; please review.

### DIFF
--- a/inst/stan/estimate-inf-and-r-rw-forecast.stan
+++ b/inst/stan/estimate-inf-and-r-rw-forecast.stan
@@ -36,6 +36,7 @@ model {
 }
 generated quantities {
   array[h] real forecast;
+  array[h] real forecast_R;
   if (h > 0) {
     array[n + h - 1] real f_rw_noise;
     for (i in 1:(n-1)) {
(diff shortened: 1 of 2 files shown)


seabbs (commented):
LGTM

seabbs:
Confirming this fixes the docs issue

sbfnk (approved):

PR #677, opened 2026-07-07
preview the wis in the evaluation session

kathsherratt:
We currently introduce the WIS in the ensembles session. It might be nice to briefly preview this earlier, in the evaluation session when we intro the CRPS (so that people who skip ensembles still see the option). 

Suggesting to add as a brief optional note in the evaluation session. For starters I just copied the existing WIS description over (so it is repetitive, but since they are in different sessions maybe OK); edits welcome.

[opening the PR here as the 'canonical' repo?]

### DIFF
--- a/sessions/forecast-evaluation.qmd
+++ b/sessions/forecast-evaluation.qmd
@@ -258,7 +258,7 @@ What does this tell you about the model?
 :::
 
 ::: {.callout-note .optional collapse="true"}
-# Probabilistic scoring without samples
+## Probabilistic scoring without samples
 Not all probabilistic forecasting models produce samples.
 The **weighted interval score (WIS)** [@bracherEvaluatingEpidemicForecasts2021] is a proper scoring rule for quantile forecasts that approximates the CRPS, by considering a weighted sum of multiple prediction intervals. 
 As the number of intervals increases, the WIS converges to the CRPS, combining sharpness and penalties for over- and under-prediction.
--- a/sessions/forecast-evaluation.qmd
+++ b/sessions/forecast-evaluation.qmd
@@ -261,7 +261,7 @@ What does this tell you about the model?
 ## Probabilistic scoring without samples
 Not all probabilistic forecasting models produce samples.
 The **weighted interval score (WIS)** [@bracherEvaluatingEpidemicForecasts2021] is a proper scoring rule for quantile forecasts that approximates the CRPS, by considering a weighted sum of multiple prediction intervals. 
-As the number of intervals increases, the WIS converges to the CRPS, combining sharpness and penalties for over- and under-prediction.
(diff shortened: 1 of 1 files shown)


sessions/forecast-evaluation.qmd
```diff
@@ -257,6 +257,13 @@ What does this tell you about the model?
   - However, the CRPS increasing with forecast horizon is a sign that the model is struggling to capture the longer term dynamics of the epidemic.
 :::
 
+::: {.callout-note .optional collapse="true"}
+# Probabilistic scoring without samples
```
sbfnk:
```suggestion
## Probabilistic scoring without samples
```
I think

sessions/forecast-evaluation.qmd
```diff
@@ -257,6 +257,13 @@ What does this tell you about the model?
...
+Not all probabilistic forecasting models produce samples.
+The **weighted interval score (WIS)** [@bracherEvaluatingEpidemicForecasts2021] is a proper scoring rule for quantile forecasts that approximates the CRPS, by considering a weighted sum of multiple prediction intervals. 
+As the number of intervals increases, the WIS converges to the CRPS, combining sharpness and penalties for over- and under-prediction.
```
sbfnk: I think giving this detail without the equation might be confusing. How about just:
```suggestion
This makes it the standard choice when forecasts are shared as quantiles or prediction intervals rather than samples, as is often done in collaborative forecasting hubs [@reichCollaborativeHubsMaking2022].
```

sbfnk (approved):

seabbs:
Personally I would prefer to do this when we first use it to avoid overloading earlier concepts but I don't feel that strongly

sessions/forecast-evaluation.qmd
```diff
@@ -257,6 +257,13 @@ What does this tell you about the model?
   - However, the CRPS increasing with forecast horizon is a sign that the model is struggling to capture the longer term dynamics of the epidemic.
 :::
 
+::: {.callout-note .optional collapse="true"}
+## Probabilistic scoring without samples
+Not all probabilistic forecasting models produce samples.
+The **weighted interval score (WIS)** [@bracherEvaluatingEpidemicForecasts2021] is a proper scoring rule for quantile forecasts that approximates the CRPS, by considering a weighted sum of multiple prediction intervals. 
+This makes it the standard choice when forecasts are shared as quantiles or prediction intervals rather than samples, as is often done in collaborative forecasting hubs [@reichCollaborativeHubsMaking2022].
+:::
```
sbfnk, line 265:
```suggestion
We will get back to the WIS in the session on [Forecast ensembles](forecast-ensembles).
:::
```

sbfnk:
as a middle ground we could move this to "going further"

PR #678, opened 2026-07-07
Short note to motivate probabilistic scoring

kathsherratt:
As the probabilistic scores we use in evaluation are mostly unfamiliar, I thought it might be helpful to add an introductory note that explicitly makes the connection from more familiar point metrics.

### DIFF
--- a/sessions/forecast-evaluation.qmd
+++ b/sessions/forecast-evaluation.qmd
@@ -63,7 +63,7 @@ One option is to evaluate a single point prediction against a single observation
 $$
 \text{error} = \text{prediction} - \text{observation}
 $$
-You may have seen these kinds of metrics: for example, mean absolute error, percentage error, root mean square error...
+You may have seen these kinds of metrics: such as mean absolute error, mean percentage error, or root mean square error.
 However, as in the rest of this course, we're focusing on probabilistic forecasts.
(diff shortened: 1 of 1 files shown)


sessions/forecast-evaluation.qmd
```diff
@@ -58,6 +58,10 @@ options(cmdstanr_print_line_numbers = TRUE)
 # Introduction to forecast evaluation
 
 An important aspect of making forecasts is that we can later confront the forecasts with what really happened and use this to assess whether our forecast model makes good predictions, or which of multiple models work best in which situation.
+One option is to evaluate a single point prediction against a single observation: error = point prediction - observation.
```
sbfnk:
```suggestion
One option is to evaluate a single point prediction against a single observation:

$$
\text{error} = \text{prediction} - \text{observation}
$$
```

sessions/forecast-evaluation.qmd
```diff
@@ -58,6 +58,10 @@ options(cmdstanr_print_line_numbers = TRUE)
...
+You may have seen these kinds of metrics: for example, mean absolute error, percentage error, root mean square error...
```
sbfnk:
```suggestion
You may have seen these kinds of metrics: such as mean absolute error, mean percentage error, or root mean square error.
```

sbfnk (approved):

